PR #90, opened 2024-11-06
Ed 4311 npm package

StoyanD:
This PR adds github actions for packaging and publishing an NPM package on tag releases. It also refactors some of the deploy code.

To trigger a deploy, tag a release : vx.x.x(-beta)
The tag will be propagated down to the npm package as its version, and it will also be set as the solidity file version that will be deployed, so use proper semver notation and ensure that the npm package is versioned correctly.

For the deploy, since we use create2; the salt is set to that tag's commit hash in the github action.


johnwhitton (approved):
Approving this now with feedback below, please read and address as needed.

**Feedback**
- [ ] NPM Package name should be eco or eco-contracts or eco-routes
- [ ] NPM Organization should change to eco-foundation or the-eco-foundation
- [ ] Build should include typechain/types and also import them
- [ ] Do we want to publish this to a github repository as part of the publish process
- [ ] ci.ts
  - [ ] const mainnetdep could be read from config rather than hardcoded
  - [ ] could use commands directly rather than needing to add them to package.json
  - [ ] Would be worth separating mainnet and sepolia deploys so can deploy each separately
  - [ ] Could add preprod environment (bigger architectural discussion)
- [ ] deployCI.ts - looks good
- [ ] deployProtocol.ts
  - [ ] Can use a nonceManager see logic below
  - [ ] Should move pre into a network rather than an option (basically we should never deploy a pre environmnet to Sepolia only Prod) bigger architecture 
- [ ] Addresses.ts
  - [ ] recommend removing `pre` and creating an environment structure based on the settlement Chain (Mainnet, Mainnet-PreProd and Sepolia)

**Future changes for ECO-BETA**
- [ ] Chain Configurations will be different for each Prover
- [ ] Will be introducing wrappers for prover contract
- [ ] Will need to leverage addresses in config for SDK
- [ ] BETA code is [here](https://github.com/the-eco-foundation/eco-routes/tree/ED-4357)
- [ ] ECO-SDK is [here](https://github.com/the-eco-foundation/eco-sdk) will need typechain types
- [ ] 


**Nonce Manager Logic**
```typescript
# Imported a nonce Manager
import { BigNumberish, BytesLike, NonceManager, toQuantity } from 'ethers'

# Created the nonce manager
  const optimismIntentCreatorNonceManager = new NonceManager(
    s.optimismIntentCreator,
  )

# Incremented it after each transaction
  await approvalTx.wait()
  optimismIntentCreatorNonceManager.increment()
```

StoyanD:
changed to eco
change org to eco-foundation
in different version
not really, or in different version
ci-
 -wanted to make the commands the same as whats used locally for deploys, next version will get rid of hardhat/ethers all together
deployprotocol- figured out how to just wait till functions returned something that makes sense
 -i want pre networks on sepolia for local dev, the other is for staging
addresses- yea probably going to modify that but i want to try integrating it on the solver and testing some versions before i figure out what works best
@johnwhitton

PR #109, opened 2024-12-04
Viem mf

StoyanD:
deploy the entire protocol to multiple networks, package the abi, sol files, and generate a ts export for the addresses that are deployed. also do verification via etherscan


contracts/Inbox.sol
```diff
@@ -55,7 +55,7 @@ contract Inbox is IInbox, Ownable {
         }
     }
 
-    function version() external pure returns (string memory) { return "v0.0.3-beta"; }
+    function version() external pure returns (string memory) { return "v0.0.4211-beta"; }
```
nknavkal, line 58: for simplicity i think we should create a version contract that is inherited by evertything else - that way we only have to update one contract, and it doesnt muddy up the code. wont affect this approval but yeah

scripts/publish/prepack.sh
```diff
@@ -25,8 +25,8 @@ rsync -av --include '*/*.sol' --exclude 'test' --exclude 'build' --exclude  'REA
 rm build/src/abi/contracts/*.dbg.json
 rm build/src/abi/interfaces/*.dbg.json
 echo "Start abi export and package.json generation"
-npx ts-node scripts/publish/abi-export.ts
```
nknavkal, line 28: is there a reason we made this switch? from a brief search it looks like tsx, while faster, does not chekc types

scripts/viem_deploy/contracts/index.ts
nknavkal, line 1: empty file, delete

nknavkal (approved):
lgtm. Tbh I'm not very sure how to review this effectively, having a tough time putting it all together, but if it works then it works. Maybe sometime we can go over how to run a deploy using this

nknavkal:
looks like there's some lint issues though can you patch those

contracts/Inbox.sol
johnwhitton: May be a version library i.e. something like [this](https://github.com/the-eco-foundation/eco-routes/blob/ED-4357/contracts/libs/Semver.sol) Already done in BETA.
```
// SPDX-License-Identifier: UNLICENSED
pragma solidity ^0.8.28;

library Semver {
    function version() internal pure returns (string memory) {
        return "0.3.1-beta.0";
    }
}
```

scripts/publish/prepack.sh
StoyanD: ts-node keeps breaking for me, there is typechecking in there though for the publish

contracts/Inbox.sol
StoyanD: ill copy it

PR #122, opened 2024-12-12
ED-4516: Cantina Audit(14): add finality delay for bedrock

johnwhitton:


contracts/Prover.sol
```diff
@@ -336,6 +337,15 @@ contract Prover is SimpleProver {
         require(
             existingSettlementBlockProof.stateRoot == l1WorldStateRoot, "settlement chain state root not yet proved"
         );
+
+        // check that the End Batch Block timestamp is greater than the current timestamp + finality delay
+        uint256 endBatchBlockTimeStamp = _bytesToUint(RLPReader.readBytes(RLPReader.readList(rlpEncodedBlockData)[11]));
+
+        require(
```
StoyanD, line 344: y not use custom errors instead of requires?

contracts/Prover.sol
```diff
@@ -499,6 +515,9 @@ contract Prover is SimpleProver {
         FaultDisputeGameProofData memory faultDisputeGameProofData,
         bytes32 l1WorldStateRoot
     ) public {
+        require(
```
StoyanD, line 518: cleaner if u use a modifier for this

StoyanD: u should add tests for this condition

contracts/Prover.sol
StoyanD: u should test this condition

StoyanD (changes requested):
lg, can u change some requires to modifiers and add tests for the new conditions your checking for?

contracts/Prover.sol
johnwhitton: nice catch 🙏 
updated now

contracts/Prover.sol
johnwhitton: nice catch updated now 🙏

johnwhitton:
Yep have removed all requires and added a modifier 🙏

StoyanD (approved):

PR #123, opened 2024-12-12
6.3 CEI pattern, 6.4 .call native transfers

nknavkal:
6.3 CEI framework
6.4 .call for native transfers


johnwhitton (approved):
Nice work 🙏 
Couple of nits/questions
1. Could use custom errors and reverts rather than require
2. Line 365 `emit Fulfillment(_expectedHash, _sourceChainID, _claimant);` should we  move this to line 378 so we don't emit the event until after we have successfully called the targets (mailbox)?

PR #125, opened 2024-12-12
22 emit bypass

nknavkal:
after 17_zero_claimant

0xweiss said i should use a nonreentrant modifier here, but personally i dont feel this is necessary since reentering this function is not beneficial and doesnt grief anyone. moved the emit above the external call though.


johnwhitton (approved):
Nice work 🙏 

Question is there any chance and what would happen if we were unable to transfer reward tokens in Intent creation (lines 96-99)

```
        for (uint256 i = 0; i < len; i++) {
            IERC20(_rewardTokens[i]).safeTransferFrom(msg.sender, address(this), _rewardAmounts[i]);
        }
        ```

PR #147, opened 2025-01-16
feat(IntentSource): gas-less intents

re1ro:
[ED-4715]

* [x] `IntentSource`: gas-less intent origination

[ED-4715]: https://eco.atlassian.net/browse/ED-4715?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/contracts/IntentFunder.sol
+++ b/contracts/IntentFunder.sol
@@ -36,6 +36,14 @@ contract IntentFunder {
 
         // Get the address that is providing the tokens for funding
         address fundingSource = intentSource.getFundingSource();
+        address refundToken = intentSource.getRefundToken();
+
+        if (refundToken != address(0)) {
+            IERC20(refundToken).safeTransfer(
+                reward.creator,
+                IERC20(refundToken).balanceOf(address(this))
+            );
+        }
 
         // Iterate through each token in the reward structure
         for (uint256 i; i < rewardsLength; ++i) {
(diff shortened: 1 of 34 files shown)


contracts/IntentFunder.sol
```diff
@@ -0,0 +1,84 @@
...
+                address(this)
+            );
+
+            // Calculate how many more tokens the vault needs to be fully funded
+            // Cast to int256 to handle the case where vault is already overfunded
+            int256 balanceDeficit = int256(amount) -
+                int256(IERC20(token).balanceOf(vault));
+
+            // Only proceed if vault needs more tokens and we have permission to transfer them
+            if (balanceDeficit > 0 && allowance > 0) {
```
StoyanD, line 58: do we need to check here? im wondering if a partially funded vault is an actual protocol state, or if the user has to somehow fund that vault before calling the intentsource

StoyanD (commented):
looks good, do we need to check how much a vault has in balance though, shouldnt the vaults be empty?

contracts/IntentFunder.sol
```diff
@@ -0,0 +1,84 @@
...
+     */
+    constructor(address vault, Reward memory reward) {
+        // Cast msg.sender to IIntentSource since we know it must be the IntentSource contract
+        IIntentSource intentSource = IIntentSource(msg.sender);
+
+        // Cache array length to save gas in loop
+        uint256 rewardsLength = reward.tokens.length;
+
+        // Get the address that is providing the tokens for funding
+        address fundingSource = intentSource.getFundingSource();
```
nknavkal, line 38: is this more expensive than just passing the fundingSource address as a param?

StoyanD (approved):

PR #152, opened 2025-01-23
adding build and assets to lint ignore

StoyanD:
adding peer dep if viem to npm lib
adding utils to the npm lib


dirkpage (approved):

scripts/viem_deploy/assets/utils/intent.ts
```diff
@@ -0,0 +1,245 @@
...
+  { type: 'function' }
+>
+
+/**
+ * The getIntentHash function from the IntentSource ABI
+ */
+type GetIntentHashFunction = Extract<
+  ExtractAbiFunctions<typeof IntentSourceAbi>,
+  { name: 'getIntentHash' }
+>['inputs'][number]
```
nknavkal, line 29: what is this inputs field? the inputs to getIntentHash is an Intent struct

scripts/viem_deploy/assets/utils/intent.ts
```diff
@@ -0,0 +1,245 @@
...
+    reward,
+  )[0]
+}
+
+/**
+ * Encodes the intent parameters
+ * @param intent the intent to encode
+ * @returns
+ */
+export function encodeIntent(intent: IntentType) {
```
nknavkal, line 153: confusd by this, this encodes the intent as the route hash and the reward hash?

nknavkal (commented):
couple questions, will approve after

scripts/viem_deploy/assets/utils/intent.ts
StoyanD: its the abi, since the structs are never defined as an instance variable, the only place i can get their abi is form functions that either get it as input args or as return args

scripts/viem_deploy/assets/utils/intent.ts
StoyanD: no, it encodes it as a bytes

scripts/viem_deploy/assets/utils/intent.ts
nknavkal: what does [number] mean here

nknavkal (approved):

PR #158, opened 2025-02-01
Introduce compressed batches to reduce message body size

carlosfebres:


contracts/prover/HyperProver.sol
```diff
@@ -82,12 +104,73 @@ contract HyperProver is IMessageRecipient, BaseProver, Semver {
...
         );
 
+        proveBatch(hashes, claimants);
+    }
+
+    /**
+     * @dev Handles the details of how to handle compressed messages.
+     * @param _messageBody The body of the message in bytes memory type.
+     */
+    function handleCompressedMessage(bytes memory _messageBody) internal {
```
re1ro, line 138: @carlosfebres isn't it the same amount of gas to call this function as just send through Hyperlane?  The gas you are saving on the source is this calldata cost. Does hyperlane charges % on top?

carlosfebres: The goal isn't to reduce the total gas consumption but to reduce the gas cost of relaying messages through Hyperlane. AFAIK they don't change anything on top (apart from the absurd gas prices they use).

re1ro: I see, that makes sense. Their gas prices make it extra expensive to relay through them

contracts/libs/MessageEncoder.sol
```diff
@@ -0,0 +1,32 @@
...
+    enum MessageType {
+        STANDARD,
+        COMPRESSED
+    }
+
+    function _getRootHash(
+        bytes32[] memory intentHashes,
+        address[] memory claimants
+    ) internal pure returns (bytes32 rootHash) {
+        bytes memory messageBody = abi.encode(intentHashes, claimants);
```
re1ro, line 16: As mentioned, we should group `hashes` by `claimant`

PR #172, opened 2025-02-13
finding 5

nknavkal:


contracts/IntentVault.sol
```diff
@@ -63,20 +63,36 @@ contract IntentVault is IIntentVault {
             // If creator is claiming, send full balance
             if (claimant == reward.creator) {
                 if (balance > 0) {
-                    IERC20(token).safeTransfer(claimant, balance);
+                    (bool success, ) = token.call(
```
re1ro, line 66: Let's use try catch here instead of low leverl call. Just for better readability and not have to hardcode `"transfer(address,uint256)"`

re1ro (changes requested):

contracts/IntentVault.sol
nknavkal: try catch doesn't like safeTransfer, getting 'Try can only be used with external function calls and contract creation calls'. it works with transfer but that's not gonna work.

re1ro: We should do try catch with `transfer` not `safeTransfer`

nknavkal: hm, okay. I was thinking this wouldnt work since transfer can fail silently, but i guess thats fine in this case since thats what call would do anyway

re1ro (approved):

PR #180, opened 2025-02-17
finding 17 & 27: removed permit calls

re1ro:


nknavkal (changes requested):
PermitCallFailed is not emitted

nknavkal (approved):

PR #183, opened 2025-02-18
finding 20: vault is now also a funder

re1ro:


contracts/IntentSource.sol
```diff
@@ -313,52 +315,92 @@ contract IntentSource is IIntentSource, Semver {
...
         bytes32 intentHash = keccak256(abi.encodePacked(routeHash, rewardHash));
 
-        if (token != address(0)) {
-            refundToken = token;
+        VaultState memory state = vaults[intentHash].state;
+
+        if (
+            state.status != uint8(RewardStatus.Claimed) &&
+            state.status != uint8(RewardStatus.Refunded) &&
+            reward.nativeValue > 0
```
nknavkal, line 375: why is this a requirement

contracts/IntentSource.sol
```diff
@@ -24,39 +23,37 @@ import {IntentVault} from "./IntentVault.sol";
...
-
-    address public fundingSource;
-
-    address public refundToken;
+    mapping(bytes32 intentHash => VaultStorage) public vaults;
 
     constructor() {}
 
     /**
      * @notice Retrieves claim state for a given intent hash
```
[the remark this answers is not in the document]
carlosfebres:
```suggestion
     * @notice Retrieves reward status for a given intent hash
```

contracts/IntentVault.sol
```diff
@@ -116,4 +171,85 @@ contract IntentVault is IIntentVault {
...
+
+    /**
+     * @notice Transfers tokens from funding source to vault using Permit2
+     * @param permit2 Permit2 contract to use for token transfer
+     * @param fundingSource Address that is providing the tokens for funding
+     * @param token Address of the token being transferred
+     * @param amount Amount of tokens to transfer
+     * @return remainingAmount Amount of tokens that still need to be transferred
+     */
+    function _transferFrom2(
```
carlosfebres, line 222: Probably worth being more verbose renaming to `_transferFromPermit2`?

contracts/IntentSource.sol
```diff
@@ -313,52 +315,92 @@ contract IntentSource is IIntentSource, Semver {
...
+    }
+
     /**
      * @notice Refunds rewards to the intent creator
      * @param routeHash Hash of the intent's route
      * @param reward Reward structure of the intent
      * @param token Optional token address for handling incorrect vault transfers
      */
-    function refundIntent(
+    function refundToken(
```
carlosfebres, line 358: We may need to rename this function to `recoverToken` and update the SolDoc. I believe it is used to recover funds that are not part of the intent. 

Since this is a recovery function, the `Refund` event shouldn't emitted here.

contracts/interfaces/IIntentSource.sol
```diff
@@ -211,18 +183,22 @@ interface IIntentSource is ISemver {
 
     /**
      * @notice Funds an intent with native tokens and ERC20 tokens
-     * @dev Allows for permit calls to approve token transfers
+     * @dev Security: this allows to call any contract from the IntentSource,
+     *      which can impose a risk if anything relies on IntentSource to be msg.sender
```
carlosfebres, line 187: We can remove this it's no longer true

carlosfebres (changes requested):

contracts/IntentSource.sol
```diff
@@ -114,55 +94,65 @@ contract IntentSource is IIntentSource, Semver {
...
+    ) external returns (bytes32 intentHash) {
         bytes32 rewardHash = keccak256(abi.encode(reward));
-        bytes32 intentHash = keccak256(abi.encodePacked(routeHash, rewardHash));
-
+        intentHash = keccak256(abi.encodePacked(routeHash, rewardHash));
         address vault = _getIntentVaultAddress(intentHash, routeHash, reward);
 
-        emit IntentFunded(intentHash, fundingAddress);
+        if (reward.nativeValue > 0 && vault.balance > 0) {
+            revert CannotFundNativeReward(intentHash);
```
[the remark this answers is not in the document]
re1ro: When vault self-destructs it refunds all the native it holds. Also having solver funding native value instead of using permit was the scenario that didn't make sense

contracts/IntentSource.sol
```diff
@@ -114,55 +94,65 @@ contract IntentSource is IIntentSource, Semver {
...
+        if (state.status == uint8(RewardStatus.Initial)) {
+            state.status = allowPartial
+                ? uint8(RewardStatus.PartiallyFunded)
+                : uint8(RewardStatus.Funded);
         }
 
-        fundingSource = fundingAddress;
+        state.mode = uint8(VaultMode.Fund);
+        state.allowPartialFunding = allowPartial ? 1 : 0;
+        state.isPermit2 = permit2 != address(0) ? 1 : 0;
```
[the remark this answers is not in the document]
re1ro: it's a fund function, not publish. It uses the state to communicate with the vault

carlosfebres (approved):

nknavkal (approved):

PR #186, opened 2025-02-20
Use Foundry to deploy contracts

carlosfebres:
# Automate Multi-Chain Contract Deployment & Verification using Foundry

## Description
This PR introduces **automated multi-chain contract deployment and verification**, requested by **Caldera (a partner)** to streamline the deployment process. By dynamically handling **chain-specific configurations**, this update simplifies deployments across multiple chains while ensuring smooth verification via Etherscan-compatible explorers.

### ✅ Deploy script `Deploy.s.sol`
- Uses **unique salts per contract** to ensure deterministic and correct deployments.  
- Writes deployment data to `deploy.csv` in **CSV format** (instead of ABI-encoded).  

### ✅ Multi-Chain Deployment (`MultiDeploy.sh`)
- Iterates over **configured chain IDs** from `.env`.  
- Dynamically fetches **RPC URLs, mailbox contracts, and Etherscan API keys**.  
- Deploys contracts using **Foundry (`forge script`)** for each chain.  
- Handles **missing configurations gracefully** by skipping unconfigured chains.  

### ✅ Contract Verification (`Verify.sh`)
- Reads deployed contract data from `deploy.csv`  
- Matches the correct **Etherscan API key per chain** dynamically.  
- Calls **`forge verify-contract`** with the correct constructor arguments.  
- Skips verification for missing API keys.  

## Why This Matters
🚀 **Requested by Caldera** → This removes friction in deploying and verifying contracts across multiple chains.  
📦 **Simplifies Multi-Chain Operations** → No more manual environment setup per chain.  
✅ **Reliable & Repeatable** → Ensures deployments are correctly verified, reducing potential errors.  

## Next Steps
- [x] Validate deployments across testnets.  
- [x] Confirm successful contract verification via Etherscan APIs.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -35,3 +35,6 @@ ETHERSCAN_API_KEY_10=
 # Hyperlane Mailbox Contracts (MAILBOX_${chainId}) Example: MAILBOX_10=... for Optimism
 MAILBOX_10=
 MAILBOX_137=
+
+# Gas Multiplier (Optional)
+GAS_MULTIPLIER_137=
--- a/.env.example
+++ b/.env.example
@@ -12,3 +12,26 @@ PROVER_PRIVATE_KEY=
 CLAIMANT_PRIVATE_KEY=
 SOLVER_PRIVATE_KEY=
 RECIPIENT_PRIVATE_KEY=
+
+## Foundry Deploy
+
+# Private Key
+PRIVATE_KEY=
+
+DEPLOY_FILE="out/deploy.csv"
+
+# Deployment Salt (Bytes 32)
+SALT=
+
+# Define Chain IDs to deploy to separed by spaces. Example: "10 8543"
+CHAIN_IDS=
+
+# RPC URLs (RPC_URL_${chainId}) Example: RPC_URL_10=... for Optimism
+RPC_URL_10=
+
+# Etherscan API Keys (ETHERSCAN_API_KEY_${chainId}) Example: ETHERSCAN_API_KEY_10=... for Optimism
+ETHERSCAN_API_KEY_10=
+
+# Hyperlane Mailbox Contracts (MAILBOX_${chainId}) Example: MAILBOX_10=... for Optimism
+MAILBOX_10=
+MAILBOX_137=
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -84,6 +84,13 @@ jobs:
         with:
           node-version-file: ".nvmrc"
           cache: "yarn"
+      - name: ⚡ Install Foundry
+        run: |
+          curl -L https://foundry.paradigm.xyz | bash
+          foundryup  # Updates Foundry to the latest version
+        shell: bash
+      - name: 🔍 Verify Installation
+        run: forge --version
       - name: Install dependencies
         uses: borales/actions-yarn@v5
         with:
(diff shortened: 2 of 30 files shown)


.env.example
```diff
@@ -12,3 +12,29 @@ PROVER_PRIVATE_KEY=
...
+# Private Key
+PRIVATE_KEY=
+
+DEPLOY_FILE="out/deploy.csv"
+
+# Deployment Salt (Bytes 32)
+SALT=
+
+# Define Chain IDs to deploy to separed by spaces. Example: "10 8543"
+CHAIN_IDS=
```
StoyanD, line 27: these should be comma separated, spaces are difficult to read

.github/workflows/publish.yaml
```diff
@@ -84,26 +84,58 @@ jobs:
...
-          OPTIMISM_BLOCKSCOUT_API_KEY: ${{ secrets.OPTIMISM_BLOCKSCOUT_API_KEY }}
-          BASE_SCAN_API_KEY: ${{ secrets.BASE_SCAN_API_KEY }}
-          ARBSCAN_API_KEY: ${{ secrets.ARBSCAN_API_KEY }}
-          CALDERA_SCAN_API_KEY: ${{ secrets.CALDERA_SCAN_API_KEY }}
-          MANTLE_SCAN_API_KEY: ${{ secrets.MANTLE_SCAN_API_KEY }}
-          POLYGON_SCAN_API_KEY: ${{ secrets.POLYGON_SCAN_API_KEY }}
-          #verification
-          ETHERSCAN_API_KEY: ${{ secrets.ETHERSCAN_API_KEY }}
+          PRIVATE_KEY: ${{ secrets.DEPLOYER_PRIVATE_KEY }}
+          CHAIN_IDS: "10 137 5000 8453 42161"
```
StoyanD, line 107: comma

.github/workflows/publish.yaml
```diff
@@ -84,26 +84,58 @@ jobs:
...
+          ETHERSCAN_API_KEY_8453: ${{ secrets.BASE_SCAN_API_KEY }}
+          ETHERSCAN_API_KEY_42161: ${{ secrets.ARBSCAN_API_KEY }}
+
+          # Hyperlane Mailbox addresses
+          MAILBOX_1: "0xc005dc82818d67AF737725bD4bf75435d065D239"
+          MAILBOX_10: "0xd4C1905BB1D26BC93DAC913e13CaCC278CdCC80D"
+          MAILBOX_137: "0x5d934f4e2f797775e53561bB72aca21ba36B96BB"
+          MAILBOX_5000: "0x398633D19f4371e1DB5a8EFE90468eB70B1176AA"
+          MAILBOX_8453: "0xeA87ae93Fa0019a82A727bfd3eBd1cFCa8f64f1D"
+          MAILBOX_42161: "0x979Ca5202784112f4738403dBec5D0F3B9daabB9"
```
StoyanD, line 131: this is going to get brittle quickly, does forge have some sort of json parser we can use to feed a config file into it?

contracts/tools/ICreate3Deployer.sol
```diff
@@ -0,0 +1,72 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.0;
+
+/**
+ * @title Deployer Contract
+ * @notice This contract is responsible for deploying and initializing new contracts using
+ * a deployment method, such as `CREATE2` or `CREATE3`.
```
StoyanD, line 8: update the comments since this isnt create2

scripts/Deploy.s.sol
```diff
@@ -0,0 +1,200 @@
...
+    }
+
+    function deployCreate3Deployer() internal {
+        address deployedCreate3Deployer = create2Factory.deploy(
+            CREATE3_DEPLOYER_BYTECODE,
+            bytes32(0)
+        );
+
+        require(
+            deployedCreate3Deployer == address(create3Deployer),
```
StoyanD, line 152: u dont hard code the deployer pk but u hardcode the create2  address of the create3 deployer. wont it break if u update the pk?

StoyanD: u should never need this, as create2 will revert or give u the correct address assuming a static deployer and nonce and bytecode

scripts/Deploy.s.sol
```diff
@@ -0,0 +1,200 @@
...
+
+        bytes32 INTENT_SOURCE_SALT = getContractSalt(salt, "INTENT_SOURCE");
+        bytes32 INBOX_SALT = getContractSalt(salt, "INBOX");
+        bytes32 HYPER_PROVER_SALT = getContractSalt(salt, "HYPER_PROVER");
+
+        vm.startBroadcast();
+
+        // Deploy deployer if hasn't been deployed
+        if (!isDeployed(address(create3Deployer))) {
+            deployCreate3Deployer();
```
StoyanD, line 48: set create3Deployer here

scripts/Deploy.s.sol
```diff
@@ -0,0 +1,200 @@
...
+        string contractPath;
+        bytes constructorArgs;
+        uint256 chainId;
+    }
+
+    SingletonFactory constant create2Factory =
+        SingletonFactory(0xce0042B868300000d44A59004Da54A005ffdcf9f);
+
+    // Create3Deployer
+    ICreate3Deployer constant create3Deployer =
```
StoyanD, line 31: dont set deployer here

scripts/Deploy.s.sol
```diff
@@ -0,0 +1,200 @@
...
+        ICreate3Deployer(0xC6BAd1EbAF366288dA6FB5689119eDd695a66814);
+
+    function run() external {
+        bytes32 salt = vm.envBytes32("SALT");
+        address mailbox = vm.envAddress("MAILBOX");
+        string memory deployFilePath = vm.envString("DEPLOY_FILE");
+        address deployer = vm.rememberKey(vm.envUint("PRIVATE_KEY"));
+
+        bytes32 INTENT_SOURCE_SALT = getContractSalt(salt, "INTENT_SOURCE");
```
StoyanD, line 40: where did u get this idea to use the contract name to derive the salt ? 🤔

scripts/Deploy.s.sol
```diff
@@ -0,0 +1,200 @@
...
+            INTENT_SOURCE_SALT
+        );
+
+        console.log("IntentSource :", address(intentSource));
+
+        // Inbox
+
+        // constructor(address _owner, bool _isSolvingPublic, address[] memory _solvers)
+        address[] memory solvers;
+        bytes memory inboxConstructorArgs = abi.encode(deployer, true, solvers);
```
StoyanD, line 68: we should really remove these fields from the inbox constructor since they are not used anymore @re1ro @nknavkal

scripts/Deploy.s.sol
```diff
@@ -0,0 +1,200 @@
...
+            bytecode,
+            sender,
+            salt
+        );
+
+        deployed = isDeployed(deployedContract);
+
+        if (!deployed) {
+            address justDeployedAddr = create3Deployer.deploy(bytecode, salt);
+            require(
```
StoyanD, line 172: how would this happen?

scripts/Deploy.s.sol
```diff
@@ -0,0 +1,200 @@
...
+                "Expected address does not match the deployed address"
+            );
+        }
+    }
+
+    function writeDeployFile(
+        string memory filePath,
+        VerificationData[3] memory contracts
+    ) internal {
+        for (uint256 i = 0; i < contracts.length; i++) {
```
StoyanD, line 183: this i called on a filename extension of csv, so u should write a line before the loop that is:
`Chain,Address,Name,Args`

StoyanD: nvm, it might get complicated looping in the .sh file

StoyanD (dismissed):
looks good, left a few comments.

we should probably move onto forge at this point

scripts/Deploy.s.sol
nknavkal: yep we have tickets up for this

scripts/Deploy.s.sol
carlosfebres: > u dont hard code the deployer pk but u hardcode the create2 address of the create3 deployer. wont it break if u update the pk?

No, because the create2 factory is the one deploying the create3 deployer

> u should never need this, as create2 will revert or give u the correct address assuming a static deployer and nonce and bytecode

The resulting address depends on the `Create3Deployer` bytecode; this validates that the correct bytecode is used in case it's changed in the future.

scripts/Deploy.s.sol
carlosfebres: I need to validate if it has been deployed before so I need the address before calling `deployCreate3Deployer`

scripts/Deploy.s.sol
carlosfebres: This deployment script assumes the create3Deployer contract is constant because the underlying contracts assume the same addresses across all chains.

scripts/Deploy.s.sol
carlosfebres: I came up with it 😄

scripts/Deploy.s.sol
carlosfebres: If for some reason the sender passed doesn't match the address deploying the contracts

.github/workflows/publish.yaml
carlosfebres: AFAIK, unfortunately not

.env.example
carlosfebres: Done 👍🏻

.github/workflows/publish.yaml
carlosfebres: Done 👍🏻

contracts/tools/ICreate3Deployer.sol
carlosfebres: Done

re1ro (approved):

PR #199, opened 2025-02-28
exit on deploy fail so we dont publish npm with fake contract addresses

StoyanD:


re1ro (approved):

StoyanD:
cleaning up and just including the fix https://github.com/eco/eco-routes/pull/214

PR #206, opened 2025-03-07
Feat/batcher rewards

nknavkal:


contracts/Inbox.sol
```diff
@@ -211,12 +225,22 @@ contract Inbox is IInbox, Eco7683DestinationSettler, Ownable, Semver {
...
-        bytes[] memory results = _fulfill(
+        (bytes[] memory results, uint256 remainingValue) = _fulfill(
             _route,
             _rewardHash,
             _claimant,
             _expectedHash
         );
+        
+        require(
+            remainingValue >= minBatcherReward,
```
re1ro, line 236: @nknavkal maybe could just use `address(this).balance` here and remove `remainingValue` calculation from `_fulfill`?

contracts/interfaces/IInbox.sol
```diff
@@ -12,6 +12,11 @@ import {Route} from "../types/Intent.sol";
  * Hyperlane instant/batched)
  */
 interface IInbox is ISemver {
+    struct ClaimantAndBatcherReward {
```
re1ro, line 15: `ClaimantWithReward`?

re1ro (approved):

PR #209, opened 2025-03-07
Intent compressor contract

carlosfebres:
# Sky-blue project 🌞 
## WIP
This is still a work in progress, and a couple of things are missing, like gas optimization, organizing, renamaing amount others.

# Summary
This contract reduces transaction call data and fees by compressing simple transfer intents and fulfillment.

This contract is immutable, meaning if more chains, tokens, or eco protocol contracts are needed, a new instance of the contract will have to be deployed with the new addresses.

## Compressing Intent Creations
This contract only supports one type of intent, a simple transfer intent. This type of intent takes a single token as a reward and a single token for the route, doing a `transfer` of the full token amount to a recipient.

Some values are already defined in the Compress contract, and depending on their nature, others are calculated on the fly. 

The salt has to be a random bytes32. We're using a simple algorithm to calculate it, which can be predetermined, but there isn't a security risk if it happens.

**Using this contract, the payload to create an intent is compressed from 29 storage slots ([sample tx](https://optimistic.etherscan.io/tx/0x4be63b29106dced5d4933ed7577582523b1d83d062132321b53ea9f0dcdcb41f)) to a single storage slot (bytes32) 🙂** 

The payload is a `bytes32` string derived from the following struct after using `abi.encodePacked`:
```solidity
struct EncodedIntent {
    uint8 sourceChainIndex;
    uint8 destinationChainIndex;
    // Reward token
    uint8 rewardTokenIndex;
    uint48 rewardAmount;
    // Route token
    uint8 routeTokenIndex;
    uint48 routeAmount;
    // Expiry duration
    uint24 expiryDuration;
}
```

## Compressing fulfillment

Using the same approach as publishing intents, fulfillment cannot calculate values on the fly since the data has to match the intent exactly. Values like the salt and the reward hash cannot be compressed.

The fulfill function is expected to be called using `DELEGATECALL` to access the wallet's funds.

**Using this contract, the payload to fulfill an intent is compressed from ~30 storage slots ([sample tx](https://optimistic.etherscan.io/tx/0x3e056fc634d9ddb2dbaf8f17e4a150ef523054cce330fd0826de9b1b5fac5310)) and two batched calls (including approval) to one call and three storage slot (3 bytes32) 🙂** 

The payload is a `bytes32` string derived from the following struct after using `abi.encodePacked`:
```solidity
struct EncodedFulfillment {
    uint8 sourceChainIndex;
    uint8 destinationChainIndex;
    // Reward token
    uint8 routeTokenIndex;
    uint48 routeAmount;
    // Prove type (INSTANT = 0, BATCH = 1)
    uint8 proveType;
    // Claimant
    address claimant;
}
```

### DIFF
--- a/contracts/compressor/IntentCompressor.sol
+++ b/contracts/compressor/IntentCompressor.sol
@@ -78,6 +78,15 @@ contract IntentCompressor {
         bytes32 routeHash = keccak256(abi.encode(route));
         bytes32 intentHash = keccak256(abi.encodePacked(routeHash, rewardHash));
 
+        require(
+            route.tokens.length == 1,
+            "Cannot fulfill intent multiple tokens"
+        );
+
+        // Approve route token
+        TokenAmount memory routeToken = route.tokens[0];
+        IERC20(routeToken.token).approve(address(INBOX), routeToken.amount);
+
         if (encodedFulfillment.proveType == 1) {
             return
                 INBOX.fulfillHyperBatched(
--- a/contracts/compressor/IntentCompressor.sol
+++ b/contracts/compressor/IntentCompressor.sol
@@ -1,0 +1,326 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.10;
+
+import {IERC20} from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import {SafeERC20} from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+
+import {IInbox} from "../interfaces/IInbox.sol";
+import {IIntentSource} from "../interfaces/IIntentSource.sol";
+import {Intent, Call, TokenAmount, Route, Reward} from "../types/Intent.sol";
+
+struct EncodedIntent {
+    uint8 sourceChainIndex;
+    uint8 destinationChainIndex;
+    // Reward token
+    uint8 rewardTokenIndex;
+    uint48 rewardAmount;
+    // Route token
+    uint8 routeTokenIndex;
+    uint48 routeAmount;
+    // Expiry duration
+    uint24 expiryDuration;
+}
+
+struct EncodedFulfillment {
+    uint8 sourceChainIndex;
+    uint8 destinationChainIndex;
+    // Reward token
+    uint8 routeTokenIndex;
+    uint48 routeAmount;
+    // Prove type (INSTANT = 0, BATCH = 1)
+    uint8 proveType;
+    // Claimant
+    address claimant;
+}
+
+contract IntentCompressor {
+    using SafeERC20 for IERC20;
+
+    /**
+     * @notice Thrown when the vault has insufficient token allowance for reward funding
+     * @param token The token address
+     * @param spender The spender address
+     * @param amount The amount of tokens required
+     */
+    error InsufficientTokenAllowance(
+        address token,
+        address spender,
+        uint256 amount
+    );
+
+    address public immutable PROVER;
+    IInbox public immutable INBOX;
+    IIntentSource public immutable INTENT_SOURCE;
+
+    constructor(address _intentSource, address _inbox, address _prover) {
+        INBOX = IInbox(_inbox);
+        INTENT_SOURCE = IIntentSource(_intentSource);
+        PROVER = _prover;
+    }
+
+    function fulfill(
+        bytes32 payload,
+        bytes32 rewardHash,
+        bytes32 routeSalt
+    ) external returns (bytes[] memory) {
+        EncodedFulfillment memory encodedFulfillment = decodeFulfillPayload(
+            payload
+        );
+
+        Route memory route = _constructRoute(
+            routeSalt,
+            encodedFulfillment.sourceChainIndex,
+            encodedFulfillment.destinationChainIndex,
+            encodedFulfillment.routeTokenIndex,
+            encodedFulfillment.routeAmount
+        );
+
+        bytes32 routeHash = keccak256(abi.encode(route));
+        bytes32 intentHash = keccak256(abi.encodePacked(routeHash, rewardHash));
+
+        if (encodedFulfillment.proveType == 1) {
+            return
+                INBOX.fulfillHyperBatched(
+                    route,
+                    rewardHash,
+                    encodedFulfillment.claimant,
+                    intentHash,
+                    PROVER
+                );
+        } else {
+            return
+                INBOX.fulfillHyperInstant(
+                    route,
+                    rewardHash,
+                    encodedFulfillment.claimant,
+                    intentHash,
+                    PROVER
+                );
+        }
+    }
+
+    function publishTransferIntentAndFund(
+        bytes32 payload
+    ) external returns (bytes32 intentHash) {
+        EncodedIntent memory encodedIntent = decodePublishPayload(payload);
+        Intent memory intent = _constructIntent(encodedIntent);
+
+        _fundIntent(intent);
+
+        return INTENT_SOURCE.publishAndFund(intent, false);
+    }
+
+    // ======================== Public Functions ========================
+
+    function getChainIds() public pure returns (uint16[6] memory) {
+        return [1, 10, 137, 8453, 5000, 42161];
+    }
+
+    function getTokens() public pure returns (address[15] memory) {
+        return [
+            // Ethereum
+            0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48, // USDC
+            0xdAC17F958D2ee523a2206206994597C13D831ec7, // USDT
+            // Optimism
+            0x0b2C639c533813f4Aa9D7837CAf62653d097Ff85, // USDC
+            0x94b008aA00579c1307B0EF2c499aD98a8ce58e58, // USDT
+            0x7F5c764cBc14f9669B88837ca1490cCa17c31607, // USDC.e
+            // Polygon
+            0x3c499c542cEF5E3811e1192ce70d8cC03d5c3359, // USDC
+            0xc2132D05D31c914a87C6611C10748AEb04B58e8F, // USDT
+            0x2791Bca1f2de4661ED88A30C99A7a9449Aa84174, // USDC.e
+            // Base
+            0x833589fCD6eDb6E08f4c7C32D4f71b54bdA02913, // USDC
+            0xd9aAEc86B65D86f6A7B5B1b0c42FFA531710b6CA, // USDbC
+            // Mantle
+            0x201EBa5CC46D216Ce6DC03F6a759e8E766e956aE, // USDT
+            0x09Bc4E0D864854c6aFB6eB9A9cdF58aC190D0dF9, // USDC
+            // Arbitrum
+            0xaf88d065e77c8cC2239327C5EDb3A432268e5831, // USDC
+            0xFF970A61A04b1cA14834A43f5dE4533eBDDB5CC8, // USDC.e
+            0xFd086bC7CD5C481DCC9C85ebE478A1C0b69FCbb9 // USDT
+        ];
+    }
+
+    function decodePublishPayload(
+        bytes32 payload
+    ) public pure returns (EncodedIntent memory) {
+        return
+            EncodedIntent({
+                sourceChainIndex: uint8(payload[0]), // uint8
+                destinationChainIndex: uint8(payload[1]), // uint8
+                rewardTokenIndex: uint8(payload[2]), // uint8
+                // Reads bytes 3 to 8 and converts them to uint48
+                rewardAmount: uint48(_extractUint(payload, 3, 8)), // uint48
+                routeTokenIndex: uint8(payload[9]), // uint8
+                // Reads bytes 10 to 15 and converts them to uint48
+                routeAmount: uint48(_extractUint(payload, 10, 15)), // uint48
+                expiryDuration: uint24(_extractUint(payload, 16, 18)) // uint24
+            });
+    }
+
+    function decodeFulfillPayload(
+        bytes32 payload
+    ) public pure returns (EncodedFulfillment memory) {
+        return
+            EncodedFulfillment({
+                sourceChainIndex: uint8(payload[0]), // uint8
+                destinationChainIndex: uint8(payload[1]), // uint8
+                routeTokenIndex: uint8(payload[2]), // uint8
+                // Reads bytes 10 to 15 and converts them to uint48
+                routeAmount: uint48(_extractUint(payload, 3, 8)), // uint48
+                proveType: uint8(payload[9]),
+                claimant: address(uint160(_extractUint(payload, 10, 29))) // uint48
+            });
+    }
+
+    // ======================== Internal Functions ========================
+
+    function _fundIntent(Intent memory intent) internal {
+        address funder = msg.sender;
+        TokenAmount[] memory tokens = intent.reward.tokens;
+
+        // Get vault address from intent
+        address vault = INTENT_SOURCE.intentVaultAddress(intent);
+
+        // Cache tokens length
+        uint256 rewardsLength = tokens.length;
+
+        // Iterate through each token in the reward structure
+        for (uint256 i; i < rewardsLength; ++i) {
+            // Get token address and required amount for current reward
+            address token = tokens[i].token;
+            uint256 amount = tokens[i].amount;
+            uint256 balance = IERC20(token).balanceOf(vault);
+
+            // Only proceed if vault needs more tokens and we have permission to transfer them
+            if (amount > balance) {
+                // Calculate how many more tokens the vault needs to be fully funded
+                uint256 remainingAmount = amount - balance;
+
+                // Check how many tokens this contract is allowed to transfer from funding source
+                uint256 allowance = IERC20(token).allowance(
+                    funder,
+                    address(this)
+                );
+
+                // Check if allowance is sufficient to fund intent
+                if (allowance < remainingAmount) {
+                    revert InsufficientTokenAllowance(
+                        token,
+                        funder,
+                        remainingAmount
+                    );
+                }
+
+                // Transfer tokens from funding source to vault using safe transfer
+                IERC20(token).safeTransferFrom(funder, vault, remainingAmount);
+            }
+        }
+    }
+
+    function _constructIntent(
+        EncodedIntent memory encodedIntent
+    ) internal view returns (Intent memory) {
+        Route memory route = _constructRoute(
+            _randomBytes32(),
+            encodedIntent.sourceChainIndex,
+            encodedIntent.destinationChainIndex,
+            encodedIntent.routeTokenIndex,
+            encodedIntent.routeAmount
+        );
+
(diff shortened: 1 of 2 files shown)


contracts/compressor/IntentCompressor.sol
```diff
@@ -0,0 +1,335 @@
...
+    uint8 sourceChainIndex;
+    uint8 destinationChainIndex;
+    // Reward token
+    uint8 rewardTokenIndex;
+    uint48 rewardAmount;
+    // Route token
+    uint8 routeTokenIndex;
+    uint48 routeAmount;
+    // Expiry duration
+    uint24 expiryDuration;
```
re1ro, line 21: Is it `expiry` or `duration`? Did you mean `deadline` ?

contracts/compressor/IntentCompressor.sol
```diff
@@ -0,0 +1,335 @@
...
+    // ======================== Public Functions ========================
+
+    function getChainIds() public pure returns (uint16[6] memory) {
+        return [1, 10, 137, 8453, 5000, 42161];
+    }
+
+    function getTokens() public pure returns (address[15] memory) {
+        return [
+            // Ethereum
+            0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48, // USDC
```
re1ro, line 131: Does it make sense to split tokens by chains? So token `0` is always USDC on any chain?

contracts/compressor/IntentCompressor.sol
```diff
@@ -0,0 +1,335 @@
...
+                tokens: routeTokens
+            });
+    }
+
+    // ======================== Private Functions ========================
+
+    function _randomBytes32() private view returns (bytes32) {
+        return
+            keccak256(
+                abi.encodePacked(block.timestamp, block.prevrandao, msg.sender)
```
re1ro, line 316: wanna add `block.coinbase` and maybe `tx.gasprice` into the mix?

re1ro: I think we need to come up with something better, this still doesn't allow to create 2 identical intents in the same block...

contracts/compressor/IntentCompressor.sol
```diff
@@ -0,0 +1,335 @@
...
+        });
+
+        TokenAmount[] memory rewardTokens;
+        rewardTokens[0] = rewardToken;
+
+        return
+            Reward({
+                nativeValue: 0,
+                prover: PROVER,
+                creator: msg.sender,
```
re1ro, line 268: So we this we can't open intents on someone's behalf?

contracts/compressor/IntentCompressor.sol
carlosfebres: So, to use a smaller number, instead of passing the deadline, the expiry duration is passed to calculate the deadline using `block.timestamp` + `expiryDuration`

contracts/compressor/IntentCompressor.sol
carlosfebres: Using a double mapping, these values will be written in storage, and accessing them will be more expensive. It probably looks cooler that way, but it may be overkill.

contracts/compressor/IntentCompressor.sol
carlosfebres: Not for this intent type since passing the `creator` will make the payload over 32 bytes long.

contracts/compressor/IntentCompressor.sol
carlosfebres: Hmmm, interesting. Instead, we can use a much smaller salt instead of doing it this way. We still have 11 bytes free in the compressed payload. We can reduce the salt to 8 bytes and add it to the payload

carlosfebres: I'm doing this, thanks @re1ro 🙂

re1ro: How about we add `gasleft()` into the mix and it will produce different salts for the same intents within a batch?

carlosfebres: That's a great idea, but then it is a bit unpredictable getting the resulting intent hash.

contracts/compressor/IntentCompressor.sol
re1ro: What do you think about `expiresIn`?

contracts/compressor/IntentCompressor.sol
```diff
@@ -0,0 +1,335 @@
...
+            if (amount > balance) {
+                // Calculate how many more tokens the vault needs to be fully funded
+                uint256 remainingAmount = amount - balance;
+
+                // Check how many tokens this contract is allowed to transfer from funding source
+                uint256 allowance = IERC20(token).allowance(
+                    funder,
+                    address(this)
+                );
+
```
re1ro, line 218: Please add the funder balance check as in the latest version of IntentSource/Vault. I think we should make it a library lol

contracts/compressor/IntentCompressor.sol
re1ro: Isn't it the whole point of random value to be unpredictable? 😄 
`block.timestamp, block.prevrandao` aren't very predictable either...

re1ro: Basically we just need to make sure to have all different salts within the same batch, what we have now, it will output the same salt for the whole batch, is that right?

contracts/compressor/IntentCompressor.sol
re1ro: That's why I said `Kinda like double mapping`. 
There is no use to have all the tokens in the array, they are from different chains and we will never loop through them. Thinking about gas efficiency, it doesn't make sense to create this whole array in memory just to be discarded.

I think the way it could be done is a function with switch cases:
```
getToken(uint8 chainId, uint8 tokenId) pure (returns address) {
    if (chainId == 1) {
      if (tokenId == 0) {
          return address(0xUSDC)
      }
  
    } else if (chainId == 10)
    ...
    }  else {
       revert Wrong();
    }
}
```

contracts/compressor/IntentCompressor.sol
re1ro: yeah that would be a second param, instead of using `msg.sender`. What if solver is using multicall? Should we add `tx.origin` to the mix?

Maybe what u want is a batch method like 
```
publishBatch(address creator, bytes32[] calldata intents) {}
```

contracts/compressor/IntentCompressor.sol
```diff
@@ -0,0 +1,335 @@
...
+        Intent memory intent = _constructIntent(encodedIntent);
+
+        _fundIntent(intent);
+
+        return INTENT_SOURCE.publishAndFund(intent, false);
+    }
+
+    // ======================== Public Functions ========================
+
+    function getChainIds() public pure returns (uint16[6] memory) {
```
re1ro, line 126: ChainID should probably be `uint256` instead of `uint16`. We use it in intent struct anyways and will save gas on conversions

re1ro: And here maybe also should be `if ... else if ... else` instead of an array.

contracts/compressor/IntentCompressor.sol
re1ro: Would make DevX much better having `USDC` as `0` on any chain

re1ro: Also allows us to have 256 token per chain instead of 256 tokens total globally.

contracts/compressor/IntentCompressor.sol
```diff
@@ -0,0 +1,336 @@
...
+        uint8 destinationChainIndex,
+        uint8 routeTokenIndex,
+        uint256 routeAmount
+    ) internal view returns (Route memory) {
+        address routeTokenTarget = getTokens()[routeTokenIndex];
+
+        Call memory routeCallTransfer = Call({
+            target: routeTokenTarget,
+            value: 0,
+            data: abi.encodeCall(IERC20.transfer, (msg.sender, routeAmount))
```
re1ro, line 289: should be a passed param instead of `msg.sender`

re1ro: I like `abi.encodeCall` 🔥

contracts/compressor/IntentCompressor.sol
```diff
@@ -0,0 +1,336 @@
...
+    // Reward token
+    uint8 routeTokenIndex;
+    uint48 routeAmount;
+    // Prove type (INSTANT = 0, BATCH = 1)
+    uint8 proveType;
+    // Claimant
+    address claimant;
+}
+
+contract IntentCompressor {
```
re1ro, line 38: I think it should be a library instead of contract, for gas efficiency we would build this directly into `Inbox`, `IntentSource`.

```
type PackedIntent is bytes32;

library IntentPacking {
    function decode(PackedIntent packed) pure returns (EncodedIntent memory) {
    
    }
    
    function publishAndFund(PackedIntent packed, address creator, address receiver) returns (bytes32 intentHash);
    
    // ...other intent methods?
}

File: contracts/compressor/IntentCompressor.sol [hunk 9]
```diff
@@ -0,0 +1,336 @@
...
+    function fulfill(
+        bytes32 payload,
+        bytes32 rewardHash,
+        bytes32 routeSalt
+    ) external returns (bytes[] memory) {
+        EncodedFulfillment memory encodedFulfillment = decodeFulfillPayload(
+            payload
+        );
+
+        Route memory route = _constructRoute(
```
re1ro, line 72: This will use `msg.sender` as `route.calls[0].transfer.recipient` and solver will try transfer to himself?

contracts/compressor/IntentCompressor.sol
```diff
@@ -0,0 +1,336 @@
...
+struct EncodedFulfillment {
+    uint8 sourceChainIndex;
+    uint8 destinationChainIndex;
+    // Reward token
+    uint8 routeTokenIndex;
+    uint48 routeAmount;
+    // Prove type (INSTANT = 0, BATCH = 1)
+    uint8 proveType;
+    // Claimant
+    address claimant;
```
re1ro, line 35: This should be `address recipient;` and  `claimant` should be `msg.sender`.

contracts/compressor/IntentCompressor.sol
```diff
@@ -0,0 +1,336 @@
...
+    // Salt
+    bytes8 salt;
+}
+
+struct EncodedFulfillment {
+    uint8 sourceChainIndex;
+    uint8 destinationChainIndex;
+    // Reward token
+    uint8 routeTokenIndex;
+    uint48 routeAmount;
```
re1ro, line 31: Is the assumption here that we round all token amounts to 6 decimals? 
USDC on BNB has 18 decimal 😭 
We should have the mapping for decimals for each token.

contracts/compressor/IntentCompressor.sol
re1ro: Added decimal to the function

contracts/compressor/IntentCompressor.sol
```diff
@@ -0,0 +1,336 @@
...
+    }
+
+    function _constructIntent(
+        EncodedIntent memory encodedIntent
+    ) internal view returns (Intent memory) {
+        Route memory route = _constructRoute(
+            bytes32(encodedIntent.salt),
+            encodedIntent.sourceChainIndex,
+            encodedIntent.destinationChainIndex,
+            encodedIntent.routeTokenIndex,
```
re1ro, line 241: decimal conversion

contracts/compressor/IntentCompressor.sol
```diff
@@ -0,0 +1,336 @@
...
+
+        TokenAmount[] memory rewardTokens = new TokenAmount[](1);
+        rewardTokens[0] = rewardToken;
+
+        return
+            Reward({
+                nativeValue: 0,
+                prover: PROVER,
+                creator: msg.sender,
+                tokens: rewardTokens,
```
re1ro, line 272: decimal conversion

contracts/compressor/IntentCompressor.sol
re1ro: decimal conversion for `routeAmount`

test/IntentCompressor.spec.ts
```diff
@@ -0,0 +1,182 @@
...
+  TestERC20,
+  IntentSource,
+  TestProver,
+  Inbox,
+  IntentCompressor,
+} from '../typechain-types'
+import { loadFixture } from '@nomicfoundation/hardhat-network-helpers'
+import { EncodedFulfillmentStruct, EncodedIntentStruct } from '../typechain-types/contracts/compressor/IntentCompressor'
+
+describe('Intent Compressor Test', (): void => {
```
re1ro, line 14: Should we start foundry tests? 😊 
So we don't have to rewrite it later when we migrate?

contracts/compressor/IntentCompressor.sol
```diff
@@ -0,0 +1,336 @@
...
+        INBOX = IInbox(_inbox);
+        INTENT_SOURCE = IIntentSource(_intentSource);
+        PROVER = _prover;
+    }
+
+    function fulfill(
+        bytes32 payload,
+        bytes32 rewardHash,
+        bytes32 routeSalt
+    ) external returns (bytes[] memory) {
```
re1ro, line 67: should maybe be payable? and add native transfer to the recipient to ancoded calls if `msg.value > 0` ?

contracts/compressor/IntentCompressor.sol
```diff
@@ -0,0 +1,336 @@
...
+                    encodedFulfillment.claimant,
+                    intentHash,
+                    PROVER
+                );
+        }
+    }
+
+    function publishTransferIntentAndFund(
+        bytes32 payload
+    ) external returns (bytes32 intentHash) {
```
re1ro, line 115: Could be payable and native value to the reward and route call for native transfer? Or should we use `tokenIndex == 0` for a native token instead of USDC?

test/IntentCompressor.spec.ts
re1ro: Maybe starting from the next PRs since you already wrote all the tests

test/IntentCompressor.spec.ts
```diff
@@ -0,0 +1,182 @@
...
+    ARB_USDT,
+    ARB_USDCe,
+  }
+
+
+  const createIntentData: EncodedIntentStruct = {
+    sourceChainIndex: ChainIdIndex.BASE,
+    destinationChainIndex: ChainIdIndex.OPTIMISM,
+    rewardTokenIndex: TokenIndex.BASE_USDC,
+    rewardAmount: (2n ** 48n) - 1n, // 1.01 MM
```
re1ro, line 58: isn't it `281 MM` if we have 6 decimal points?

contracts/compressor/IntentCompressor.sol
re1ro: You can bump this type as you still have room in bytes32
`uint64 routeAmount;`

contracts/compressor/IntentCompressor.sol
```diff
@@ -0,0 +1,336 @@
...
+import {IInbox} from "../interfaces/IInbox.sol";
+import {IIntentSource} from "../interfaces/IIntentSource.sol";
+import {Intent, Call, TokenAmount, Route, Reward} from "../types/Intent.sol";
+
+struct EncodedIntent {
+    uint8 sourceChainIndex;
+    uint8 destinationChainIndex;
+    // Reward token
+    uint8 rewardTokenIndex;
+    uint48 rewardAmount;
```
re1ro, line 16: `uint64 rewardAmount;`

contracts/compressor/IntentCompressor.sol
```diff
@@ -0,0 +1,336 @@
...
+    // Route token
+    uint8 routeTokenIndex;
+    uint48 routeAmount;
```
re1ro, line 19: `uint64 routeAmount;`

contracts/compressor/IntentCompressor.sol
re1ro: Since we have more room in `bytes32` we can change it to 
`uint96 deadline;`

carlosfebres: Hmmmm I like `expiresIn` better. 👍🏻 

I think we can leave it as a uint24 since that gives a maximum of 194 days to expiry duration, and I'm already using 8 bytes for the salt.

contracts/compressor/IntentCompressor.sol
carlosfebres: Before merging, let's make sure we are using all the bytes. 👌🏻 

We may need another solution because, for 18 decimals, using 64 bytes will give us a maximum of 18 units (e.x. 18.44... USDC). We could use a float approach with some precision plus an exponent. 

This approach wouldn't work for very precise amount but could work for USD tokens with 18 decimals.

contracts/compressor/IntentCompressor.sol
carlosfebres: Isn't the recipient the same as claimant in this context? 🤔

carlosfebres: We could totally constrain the recipient to always be the `msg.sender` (actually `address(this)` since this function is intended to be used with `DELEGATECALL`), but since there's enough space, I just added it here.

contracts/compressor/IntentCompressor.sol
carlosfebres: Totally, great idea! I build it this way!

contracts/compressor/IntentCompressor.sol
carlosfebres: Yeah, great idea as well 👌🏻

contracts/compressor/IntentCompressor.sol
carlosfebres: Lol. Nice catch. I haven't finished the unit tests yet 😅

contracts/compressor/IntentCompressor.sol
re1ro: This is solved here https://github.com/eco/eco-routes/pull/209#discussion_r1985958273
By having encoding decimals. Could be 9 decimals for most practical cases.

contracts/compressor/IntentCompressor.sol
re1ro: Yeah you have so much unused space left in bytes32, what's the point of introducing this limitation?
You can basically have
```
uint64 rewardAmount;
uint64 routeAmount;
uint56 deadline;
bytes8 salt;
```

contracts/compressor/IntentCompressor.sol
re1ro: On the source msg.sender is the recipient to be on destination.
But on destination, msg.sender is the solver, so solver is trying to send all the money to themselves?
This should be `address recipient;`

carlosfebres: I see what you mean now. Changing claimant for recipient. 👍🏻

PR #210, opened 2025-03-13
make inbox immutable

nknavkal:


nknavkal:
already done elsewhere

PR #212, opened 2025-03-19
license update 

shahrukhtrao:
This updates the license from MIT to BSL


nknavkal (approved):
i aint readin all that

PR #213, opened 2025-03-23
feat(metarouter): add metaprover contract

jake-nyquist:
Adds a "Metarouter" contract which sends proof messages via the Metalayer.

TODO: 
- Migrate to NPM version of Metalayer interfaces
- Verify test coverage


contracts/Eco7683DestinationSettler.sol
```diff
@@ -66,6 +66,23 @@ abstract contract Eco7683DestinationSettler is IDestinationSettler {
                 metadata,
                 postDispatchHook
             );
+        } else if (proofType == IProver.ProofType.Metalayer) {
```
nknavkal, line 69: for what its worth we're not going to be supporting 7683 any further, we haven't and wont be deploying it

nknavkal: we do still have to remove it from the implementaiton though

PR #223, opened 2025-04-16
ED-5151: messsage-based prover / inbox refactor

nknavkal:


nknavkal:
some changes made here: 
- took 7683 out completely
- getProofType now returns a string, look for 'Hyperlane' or 'storage'
- new event names, emitted from new locations

contracts/Inbox.sol
```diff
@@ -177,6 +118,18 @@ contract Inbox is IInbox, Eco7683DestinationSettler, Ownable, Semver {
...
+        hashes[0] = _expectedHash;
+        claimants[0] = _claimant;
+
+        uint256 fee = MessageBridgeProver(_localProver).fetchFee(
+            _route.source,
+            hashes,
+            claimants,
+            _sourceChainProver,
+            _data
+        );
```
carlosfebres, line 132: We may not want to fetch the fee. Instead, we can just send all the `msg.value` to the "Message Bridge Prover".

contracts/Inbox.sol
```diff
@@ -94,77 +89,23 @@ contract Inbox is IInbox, Eco7683DestinationSettler, Ownable, Semver {
...
-        uint256 fee = fetchFee(
-            _route.source,
-            _prover32,
-            messageBody,
-            _metadata,
-            _postDispatchHook
-        );
+        address _localProver,
+        address _sourceChainProver,
+        bytes calldata _data
```
carlosfebres, line 107: Can we rename `_data` to a more verbose name? Probably something like `proverContext` or `messageSenderData`.

contracts/Inbox.sol
```diff
@@ -94,77 +89,23 @@ contract Inbox is IInbox, Eco7683DestinationSettler, Ownable, Semver {
...
-
-        uint256 fee = fetchFee(
-            _route.source,
-            _prover32,
-            messageBody,
-            _metadata,
-            _postDispatchHook
-        );
+        address _localProver,
+        address _sourceChainProver,
```
carlosfebres, line 106: The `_sourceChainProver` should be included in the `_data` field since the source chain prover may be specific to our provers.

contracts/Inbox.sol
```diff
@@ -188,36 +141,42 @@ contract Inbox is IInbox, Eco7683DestinationSettler, Ownable, Semver {
...
-    function fulfillHyperBatched(
+    function fulfillMessageBridgeBatched(
         Route calldata _route,
         bytes32 _rewardHash,
         address _claimant,
         bytes32 _expectedHash,
-        address _prover
+        address _localProver,
+        address _sourceChainProver
     ) external payable returns (bytes[] memory) {
```
carlosfebres, line 172: Batched fulfillments are specific to the prover contract; this logic must be moved to the prover and specify the fulfillment type (instant fulfillments or batch fulfillments) using the prover context/data.

carlosfebres (changes requested):
We need to move the batch fulfillments logic out of the Inbox to the HyperProver. And remove all Hyperlane logic from the Inbox, like the mailbox contract address.

contracts/Inbox.sol
nknavkal: sourceChain prover is going to be like the thing that receives the message on the other side, i would imagine any msg-based system would have this right

contracts/Inbox.sol
nknavkal: hm, i kinda feel like any message-based prover should be able to do batch fulfills? is it necessary to have that be on the prover?

contracts/Inbox.sol
nknavkal: i elected to do it this way to make it easier to send the remaining funds back to sender and separate the logic nicely

contracts/Inbox.sol
carlosfebres: Probably, but that'd still be the prover's responsibility. If we want provers to easily support a batching mechanism, we can create an abstract `BaseProver` contract that facilitates that implementation.

contracts/Inbox.sol
carlosfebres: I wouldn't assume that; there may be signature-styled provers that don't need to pass a message. Also, by including this source chain prover in the prover context, we can remove parameters.

PR #226, opened 2025-04-19
Ed 5151 

nknavkal:
this should be a steady state for now. I have another idea, but its going to be a slightly larger haul i dont think we can accomplish before caldera's audit, but this should be everything.


contracts/Inbox.sol
```diff
@@ -46,11 +42,9 @@ contract Inbox is IInbox, Eco7683DestinationSettler, Ownable, Semver {
     constructor(
         address _owner,
         bool _isSolvingPublic,
-        uint96 _minBatcherReward,
         address[] memory _solvers
```
StoyanD, line 45: do we use this public/private owner anywhere or can we remove it?

PR #240, opened 2025-05-14
finding 4: Prover address mismatch between source and destination chains can lead to DoS

nknavkal:


test/DestinationSettler.spec.ts
```diff
@@ -17,7 +17,7 @@ import {
   encodeIntent,
 } from '../utils/intent'
 
-describe('Destination Settler Test', (): void => {
+describe.only('Destination Settler Test', (): void => {
```
re1ro, line 20: @nknavkal please undo only

re1ro (approved):

test/DestinationSettler.spec.ts
nknavkal: woops

PR #252, opened 2025-05-22
Caldera audit intentsource attackfix

nknavkal:


contracts/IntentSource.sol
```diff
@@ -248,19 +248,22 @@ contract IntentSource is IIntentSource, Semver {
...
     /**
      * @notice Withdraws rewards associated with an intent to its claimant
-     * @param routeHash Hash of the intent's route
-     * @param reward Reward structure of the intent
+     * @param _intent The intent to withdraw rewards for
      */
-    function withdrawRewards(bytes32 routeHash, Reward calldata reward) public {
-        bytes32 rewardHash = keccak256(abi.encode(reward));
+    function withdrawRewards(Intent calldata _intent) public {
+        bytes32 rewardHash = keccak256(abi.encode(_intent.reward));
```
re1ro, line 254: We have an internal function for hashing

contracts/IntentSource.sol
```diff
@@ -248,19 +248,22 @@ contract IntentSource is IIntentSource, Semver {
...
+        bytes32 routeHash = keccak256(abi.encode(_intent.route));
         bytes32 intentHash = keccak256(abi.encodePacked(routeHash, rewardHash));
 
-        address claimant = BaseProver(reward.prover).provenIntents(intentHash);
+        (address claimant, uint96 destinationChainID) = BaseProver(
```
re1ro, line 258: should it be in reverse order? for keeping storage compatibility

contracts/IntentSource.sol
```diff
@@ -273,45 +276,38 @@ contract IntentSource is IIntentSource, Semver {
...
-            new Vault{salt: routeHash}(intentHash, reward);
+            new Vault{salt: routeHash}(intentHash, _intent.reward);
 
             return;
         }
 
         if (claimant == address(0)) {
             revert UnauthorizedWithdrawal(intentHash);
+        } else if (destinationChainID != uint96(_intent.route.destination)) {
+            revert WrongDestinationChain(intentHash);
```
re1ro, line 287: Should we try to reset the storage in prover and save the right chain id here? Instead of reverting?

re1ro:
@nknavkal should we use a different base branch to see clean diff?

contracts/IntentSource.sol
```diff
@@ -248,14 +249,24 @@ contract IntentSource is IIntentSource, Semver {
...
-     * @param reward Reward structure of the intent
+     * @param _intent The intent to withdraw rewards for
      */
-    function withdrawRewards(bytes32 routeHash, Reward calldata reward) public {
-        bytes32 rewardHash = keccak256(abi.encode(reward));
-        bytes32 intentHash = keccak256(abi.encodePacked(routeHash, rewardHash));
+    function withdrawRewards(Intent calldata _intent) public {
+        (bytes32 intentHash, bytes32 routeHash, ) = getIntentHash(_intent);
+
+        (uint96 destinationChainID, address claimant) = BaseProver(
```
re1ro, line 257: I think we should use IProver here as well then

contracts/IntentSource.sol
```diff
@@ -7,7 +7,8 @@ import {SafeERC20} from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol
 
 import {IIntentSource} from "./interfaces/IIntentSource.sol";
 import {BaseProver} from "./prover/BaseProver.sol";
```
re1ro, line 9: remove this import then?

contracts/prover/MessageBridgeProver.sol
```diff
@@ -5,23 +5,33 @@ pragma solidity ^0.8.26;
...
  * @notice Abstract contract for cross-chain message-based proving mechanisms
+ * @notice the terms "source" and "destination" are used in reference to a given intent: created on source chain, fulfilled on destination chain
  * @dev Extends BaseProver with functionality for message bridge provers like Hyperlane and Metalayer
  */
 abstract contract MessageBridgeProver is
     BaseProver,
     IMessageBridgeProver,
     Whitelist
 {
+    uint256 public constant RARICHAIN_CHAIN_ID = 1380012617;
```
re1ro, line 21: so why are we hardcoding only this one chain?

nknavkal: this is the only chain we care about/will be deploying to in the immediate future where domainID differs from chainID

contracts/prover/HyperProver.sol
```diff
@@ -123,24 +122,31 @@ contract HyperProver is IMessageRecipient, MessageBridgeProver, Semver {
...
+            _claimants,
+            unpacked
+        );
 
         // Send the message through Hyperlane mailbox using local variables
         // Note: Some Hyperlane versions have different dispatch signatures.
         // This matches the expected signature for testing.
         IMailbox(MAILBOX).dispatch{value: fee}(
-            unpacked.sourceChainDomain,
+            destinationDomain,
```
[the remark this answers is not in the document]
nknavkal: good catch, sourceChainDomain is correct

PR #281, opened 2025-07-11
Failed chains

StoyanD:


scripts/Deploy.s.sol
```diff
@@ -28,6 +29,11 @@ contract Deploy is Script {
     SingletonFactory constant create2Factory =
         SingletonFactory(0xce0042B868300000d44A59004Da54A005ffdcf9f);
 
+    // CreateX contract address for World Chain (480)
+    address constant CREATEX_ADDRESS =
```
re1ro, line 33: since we are defining 
`ICreateX constant createXContract` at the top of the file, we might not even need a separate contant. Especially if we are not creating separate contants for `create2Factory` and `create3Deployer`

re1ro (approved):

PR #297, opened 2025-07-30
fix: address solanamainnet portal/hyperProver addresses

aroralanuk:
- Added static solana addresses bc we deployed them using anchor (not using the routes deployment)
- Added EcoAddress type for easier transform for the AddressTransform in the eco-solver


scripts/semantic-release/sr-build-package.ts
```diff
@@ -429,6 +430,11 @@ export function getContractAddress<T extends EcoChainIdsEnv>(
 ): Hex {
   return EcoProtocolAddresses[chainId][contractName] as Hex;
 }
+
+/**
+ * The multi-chain address type for the eco protocol.
+ */
+export type EcoAddress = EvmAddress | SvmAddress;
```
fish-sammy, line 437:
```suggestion
export type Address = EvmAddress | SvmAddress;
```
EcoAddress is kinda weird.

scripts/semantic-release/sr-deploy-contracts.ts
```diff
@@ -149,6 +152,23 @@ async function generateDeploymentAddressesJSON(
...
+function staticSvmAddresses(): Record<string, Record<string, string>> {
+  return {
+    '1399811150-pre': {
+      IntentSource: '64Xrmg8iLpvW6ohBcjubTqXe56iNYqRi52yrnMfnbaA6',
+      Inbox: '64Xrmg8iLpvW6ohBcjubTqXe56iNYqRi52yrnMfnbaA6',
+      HyperProver: '8bvpmgp9xbGngm9KmfX5poJer8dW4BJb7LaxuSmfCPZz',
+      MetaProver: '0x0000000000000000000000000000000000000000',
+    },
+  }
+}
```
fish-sammy, line 170: Should just store in a JSON file and import in since it's static.

scripts/semantic-release/sr-deploy-contracts.ts
```diff
@@ -132,7 +132,10 @@ async function generateDeploymentAddressesJSON(
   logger.log('Creating the deployAddresses.json...')
 
   try {
```
fish-sammy, line 134: Not related to this PR, but I'd use import { writeFile } from 'fs/promises'; so that u can just do

    const contractsJson = {
      ...processContractsForJson(contracts),
      ...staticSvmAddresses(),
    }
    // Save to deployed addresses JSON
    const deployedAddressesPath = getDeployedAddressesJsonPath(cwd)

    writeFile(deployedAddressesPath, JSON.stringify(contractsJson, null, 2))
      .then(() => {
        logger.log(`Contract addresses saved to ${deployedAddressesPath}`)
        logger.log('✅ Contract deployment completed successfully')
      })
      .catch((err) => {
        logger.error(`Deployment process failed: ${(error as Error).message}`);

        throw err;
      })
Also, writeFileSync is blocking.

PR #301, opened 2025-08-01
feat: add generic `RewardType` and `RouteType` to support solana

aroralanuk:


scripts/semantic-release/assets/utils/intent.ts
```diff
@@ -130,10 +195,27 @@ export type RewardType = IntentType['reward']
  * });
  */
 export function encodeRoute(route: RouteType) {
```
fish-sammy, line 197:
```suggestion
export function encodeRoute(route: RouteType): Hex {
```

scripts/semantic-release/assets/utils/intent.ts
```diff
@@ -173,11 +255,29 @@ export function decodeRoute(route: Hex): RouteType {
...
+      // using anchor's BorshCoder for synchronous encoding
+      const { deadline, creator, prover, nativeAmount, tokens } = reward
+      const encoded = svmCoder.types.encode('reward', {
+        deadline,
+        creator,
+        prover,
+        nativeAmount,
+        tokens,
+      })
+      console.log('SVM encoded', encoded)
```
fish-sammy, line 275:
```suggestion
```

scripts/semantic-release/tests/intents.spec.ts
```diff
@@ -0,0 +1,152 @@
...
+        ],
+      }
+
+      const encoded = encodeReward(evmReward)
+      const decoded = decodeReward(encoded)
+
+      expect(decoded).toEqual(evmReward)
+    })
+  })
+})
```
fish-sammy, line 152:
```suggestion
})

```

scripts/semantic-release/tests/intents.spec.ts
```diff
@@ -0,0 +1,152 @@
...
+            amount: 500000000000000000n,
+          },
+          {
+            token: '0xB1c86a33E6441e7c34A88A39bEAD93bB31d4fc6F' as `0x${string}`,
+            amount: 250000000000000000n,
+          },
+        ],
+      }
+
+      const encoded = encodeReward(evmReward)
```
fish-sammy, line 32: Should add `expect(encoded).toMatchSnapshot();` and all the other tests in this `describe`. The issue right now is that, if we update our encoding/decoding dependencies, chances are these tests are still gonna pass but changes to the `encoded` may be unexpected. We need the tests to catch that kinda change and decide if the external libs are buggy and if the contracts match the versioning.

fish-sammy (changes requested):

PR #302, opened 2025-08-01
fix: fulfill with executor instead of inbox itself as the msg.sender

fish-sammy:
https://cantina.xyz/code/9fd0c3da-dd38-409f-8a8a-9e34da467509/findings?order_by=severity&severity=critical,high,medium&finding=1

### DIFF
--- a/contracts/Executor.sol
+++ b/contracts/Executor.sol
@@ -1,0 +1,109 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.26;
+
+import {IERC165} from "@openzeppelin/contracts/utils/introspection/IERC165.sol";
+
+import {IProver} from "./interfaces/IProver.sol";
+import {IExecutor} from "./interfaces/IExecutor.sol";
+
+import {Call} from "./types/Intent.sol";
+
+/**
+ * @title Executor
+ * @notice Contract for secure execution of intent calls
+ * @dev Implements IExecutor with safety checks to prevent malicious calls
+ */
+contract Executor is IExecutor {
+    /**
+     * @notice Interface ID for IProver used to detect prover contracts
+     */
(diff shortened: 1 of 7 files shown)


contracts/Executor.sol
```diff
@@ -0,0 +1,109 @@
...
+contract Executor is IExecutor {
+    /**
+     * @notice Interface ID for IProver used to detect prover contracts
+     */
+    bytes4 private constant IPROVER_INTERFACE_ID = type(IProver).interfaceId;
+
+    /**
+     * @notice Address of the portal contract authorized to call execute
+     */
+    address private portal;
```
re1ro, line 25: immutable?

contracts/Executor.sol
```diff
@@ -0,0 +1,89 @@
...
+     */
+    constructor() {
+        portal = msg.sender;
+    }
+
+    /**
+     * @notice Restricts function access to the portal contract only
+     * @dev Reverts with Unauthorized error if caller is not the portal
+     */
+    modifier onlyPortal() {
```
nknavkal, line 34: nitpick but this is only used once, could be a require rather than a modifier

fish-sammy: I find it more readable with a modifier but don't have strong opinion.

contracts/Inbox.sol
```diff
@@ -22,22 +23,21 @@ import {DestinationSettler} from "./ERC7683/DestinationSettler.sol";
...
-     */
-    bytes4 public constant IPROVER_INTERFACE_ID = type(IProver).interfaceId;
-
     /**
      * @notice Mapping of intent hashes to their claimant identifiers
      * @dev Stores the cross-VM compatible claimant identifier for each fulfilled intent
      */
     mapping(bytes32 => bytes32) public fulfilled;
 
+    IExecutor public executor;
```
hjorthjort, line 32: `immutable` would save you some gass

PR #305, opened 2025-08-07
feat: integrate Vault V2 in the IntentSource

fish-sammy:


contracts/IntentSource.sol
```diff
@@ -51,28 +64,8 @@ abstract contract IntentSource is OriginSettler, IVaultStorage, IIntentSource {
...
-    }
-
-    /**
-     * @notice Retrieves the permitContract address funding an intent
-     */
-    function getPermitContract(
-        bytes32 intentHash
-    ) external view virtual returns (address) {
-        return vaults[intentHash].permitContract;
+    ) public view override returns (IVaultV2.Status status) {
```
nknavkal, line 67: why is this override?

contracts/interfaces/IIntentSource.sol
```diff
@@ -354,7 +285,7 @@ interface IIntentSource is IVaultStorage {
         address funder,
         address permitContract,
         bool allowPartial
-    ) external returns (bytes32 intentHash, address vault);
```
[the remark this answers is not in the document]
fish-sammy: cuz this guy @re1ro insists lol. I can explain in a call, hard to do it by typing.

contracts/IntentSource.sol
```diff
@@ -114,10 +107,35 @@ abstract contract IntentSource is OriginSettler, IVaultStorage, IIntentSource {
     )
         public
         pure
-        virtual
+        override
```
nknavkal, line 110: not really understanding why all of these are override, or honestly why they were virtual to begin with

fish-sammy: `virtual` is obviously wrong. `override` because there functions are defind in the interface `IIntentSource`

contracts/IntentSource.sol
nknavkal: might be a philosophical thing here, i think we had talked ab this with kiryl earlier, but i usually only override if there is an implementation this supplants. Otherwise the question 'what does this override' doesnt really have an answer. I think this is the convention we have right now everywhere else

contracts/IntentSource.sol
```diff
@@ -151,29 +167,18 @@ abstract contract IntentSource is OriginSettler, IVaultStorage, IIntentSource {
...
-        override(IIntentSource, OriginSettler)
-        returns (address)
-    {
-        (bytes32 intentHash, bytes32 routeHash, ) = getIntentHash(
-            destination,
-            route,
-            reward
-        );
-        return _getIntentVaultAddress(intentHash, routeHash, reward);
+    ) public view override(IIntentSource, OriginSettler) returns (address) {
```
nknavkal, line 170: does it work if you dont do this specificier? is the method on the interface marked as virtual?

contracts/interfaces/IVaultV2.sol
```diff
@@ -20,6 +20,12 @@ interface IVaultV2 {
     /// @notice Thrown when vault status is invalid for withdrawal operation
     error InvalidStatusForWithdrawal(Status status);
 
+    /// @notice Thrown when attempting to recover an invalid token (zero address or reward token)
+    error InvalidRecoverToken(address token);
+
+    /// @notice Thrown when attempting to recover a token with zero balance
+    error ZeroRecoverTokenBalance(address token);
+
     /// @notice Thrown when vault status is invalid for refund operation or deadline not reached
```
nknavkal, line 29: not changed in this PR but InvalidClaimant seems to only be invoked when claimant is zero. can call it zeroclaimant and omit the argument

nknavkal (commented):
looks mostly good, couple questions

contracts/IntentSource.sol
fish-sammy: 1. No it's required when implementing the same methods from multiple bases 2. Methods in interfaces cannot be marked virtual

contracts/IntentSource.sol
```diff
@@ -737,150 +542,128 @@ abstract contract IntentSource is OriginSettler, IVaultStorage, IIntentSource {
...
+                _fundToken(vault, token, reward.tokens[i].amount);
         }
 
-        // Update vault state based on funding result
-        vaults[intentHash].state.status = partiallyFunded
-            ? uint8(RewardStatus.PartiallyFunded)
-            : uint8(RewardStatus.Funded);
-        emit IntentFunded(intentHash, funder, !partiallyFunded);
+        if (!allowPartial && !funded) {
+            revert InsufficientFunds(intentHash);
```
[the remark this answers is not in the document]
fish-sammy: Yes it would. `fundToken` pulls whatever is available up to what's required based on reward requirement and existing balance. The `allowPartial` just happens in the end.

nknavkal (approved):
should be good

PR #237, opened 2025-05-09
feat: add Universal Intent Source for Cross-Chain Compatibility

re1ro:
## Summary
- Add cross-chain compatibility through a dual-type system supporting both EVM and non-EVM chains

PR #315, opened 2025-08-13
fix: address comments in #237

fish-sammy:
Address comments in https://github.com/eco/eco-routes/pull/237

### DIFF
--- a/contracts/interfaces/IProver.sol
+++ b/contracts/interfaces/IProver.sol
@@ -31,6 +31,12 @@ interface IProver is ISemver {
     error ZeroPortal();
 
     /**
+     * @notice Chain ID is too large to fit in uint64
+     * @param chainId The chain ID that is too large
+     */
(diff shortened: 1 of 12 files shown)


contracts/interfaces/IProver.sol
```diff
@@ -38,6 +43,12 @@ interface IProver is ISemver {
         uint64 destination
     );
 
+    /**
+     * @notice Emitted when an intent proof is challenged
+     * @param intentHash Hash of the challenged intent
+     */
+    event IntentProofChallenged(bytes32 indexed intentHash);
```
nknavkal, line 50: seeing where this is called, this event represents deleting an invalid intent proof, 'challenged' leaves a question of whether or not the challenge was successful. IntentProofInvalidated? Deleted?

fish-sammy: I like invalidated.

PR #319, opened 2025-08-13
feat: add `PolymerProver`

i-norden:
`PolymerProver` contract that implements the `IProver` interface on-top of [CrossL2ProverV2.sol](https://github.com/polymerdao/prover-contracts/blob/main/contracts/core/prove_api/CrossL2ProverV2.sol).

1. Solver fulfills intent on destination chain by calling `Inbox.fulfillAndProve()`
2. `IProver(PolymerProver).prove()` is called and emits `IntentFulfilledFromSource` events
3. Solver calls [Prove API](https://docs.polymerlabs.org/docs/build/prove-api-V2/api-endpoints) for proof of log containing the event
4. `PolymerProver.validate()` is called on source chain with the proof returned in 3 as the argument
5. PolymerProver calls `CrossL2ProverV2.validateEvent()` to validate the event proof and extract `intentHash` and `ProofData`
7. `BaseProver._provenIntents` mapping is updated with intentHash->ProofData
8. Solver can now call `IntentSource.withdraw()` on the source chain to withdraw rewards


contracts/prover/PolymerProver.sol
```diff
@@ -0,0 +1,242 @@
...
+
+    /**
+     * @notice Validates multiple proofs in a batch
+     * @param proofs Array of proof data to validate
+     */
+    function validateBatch(bytes[] calldata proofs) external {
+        for (uint256 i = 0; i < proofs.length; i++) {
+            (bytes32 intentHash, address claimant, uint64 destinationChainId) = _validateProof(proofs[i]);
+            processIntent(intentHash, claimant, destinationChainId);
+        }
```
fish-sammy, line 90:
```suggestion
        for (uint256 i = 0; i < proofs.length; i++) {
            validate(proofs[i])
        }
```
and change `validate` from `external` to `public`.

contracts/prover/PolymerProver.sol
```diff
@@ -0,0 +1,242 @@
...
+            uint32 destinationChainId,
+            address emittingContract,
+            bytes memory topics,
+            /* bytes memory data */
+        ) = CROSS_L2_PROVER_V2.validateEvent(proof);
+
+        if (!isWhitelisted(uint64(destinationChainId), bytes32(uint256(uint160(emittingContract))))) {
+            revert InvalidEmittingContract(emittingContract);
+        }
+        checkTopicLength(topics, 128); // 4 topics: signature + hash + claimant + sourceChainId
```
fish-sammy, line 115: I'd just inline this and remove the function. It's a one-liner anyway. Also, create a immutable constant for `128` please.

contracts/prover/PolymerProver.sol
```diff
@@ -0,0 +1,242 @@
...
+            let topicsPtr := add(topics, 32)
+            let arrayPtr := add(topicsArray, 32)
+
+            mstore(arrayPtr, mload(topicsPtr))
+            mstore(add(arrayPtr, 32), mload(add(topicsPtr, 32)))
+            mstore(add(arrayPtr, 64), mload(add(topicsPtr, 64)))
+            mstore(add(arrayPtr, 96), mload(add(topicsPtr, 96)))
+        }
+
+        checkTopicSignature(topicsArray[0], PROOF_SELECTOR);
```
fish-sammy, line 129: Same here.

contracts/prover/PolymerProver.sol
```diff
@@ -0,0 +1,242 @@
...
+        // Convert bytes32 claimant to address
+        claimant = address(uint160(uint256(topicsArray[2])));
```
fish-sammy, line 131: Use `AddressConverter`.

contracts/prover/PolymerProver.sol
```diff
@@ -0,0 +1,242 @@
...
+            bytes memory topics,
+            /* bytes memory data */
+        ) = CROSS_L2_PROVER_V2.validateEvent(proof);
+
+        if (!isWhitelisted(uint64(destinationChainId), bytes32(uint256(uint160(emittingContract))))) {
+            revert InvalidEmittingContract(emittingContract);
+        }
+        checkTopicLength(topics, 128); // 4 topics: signature + hash + claimant + sourceChainId
+
+        bytes32[] memory topicsArray = new bytes32[](4);
```
fish-sammy, line 117: I feel it's more readable to create 4 bytes32 variables than having them in an array. Don't have to rememeber what each index is for reading the code after that way.

contracts/prover/PolymerProver.sol
```diff
@@ -0,0 +1,242 @@
...
+            mstore(add(arrayPtr, 64), mload(add(topicsPtr, 64)))
+            mstore(add(arrayPtr, 96), mload(add(topicsPtr, 96)))
+        }
+
+        checkTopicSignature(topicsArray[0], PROOF_SELECTOR);
+        // Convert bytes32 claimant to address
+        claimant = address(uint160(uint256(topicsArray[2])));
+        // Get sourceChainId from event topic and verify it matches current chain
+        uint64 eventSourceChainId = uint64(uint256(topicsArray[3]));
+        if (eventSourceChainId != block.chainid) revert InvalidSourceChain();
```
fish-sammy, line 134: So polymer's domain ID equals to the chain ID?

contracts/prover/PolymerProver.sol
```diff
@@ -0,0 +1,242 @@
...
+     */
+    function processIntent(bytes32 intentHash, address claimant, uint64 destination) internal {
+        ProofData storage proof = _provenIntents[intentHash];
+        if (proof.claimant != address(0)) {
+            emit IntentAlreadyProven(intentHash);
+        } else {
+            proof.claimant = claimant;
+            proof.destination = destination;
+            emit IntentProven(intentHash, claimant, destination);
+        }
```
fish-sammy, line 154:
```suggestion
        if (proof.claimant != address(0)) {
            emit IntentAlreadyProven(intentHash);

            return;
        }


        proof.claimant = claimant;
        proof.destination = destination;

        emit IntentProven(intentHash, claimant, destination);
```
Not a fan of if-else.

contracts/prover/PolymerProver.sol
```diff
@@ -0,0 +1,242 @@
...
+    }
+
```
fish-sammy, line 156: What's this function for? Who's calling it?

fish-sammy (changes requested):

contracts/interfaces/ICrossL2ProverV2.sol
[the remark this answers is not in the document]
i-norden: Correct

contracts/prover/PolymerProver.sol
```diff
@@ -0,0 +1,277 @@
...
+        for (uint256 i = 0; i < numPairs; i++) {
+            uint256 offset = i * 64;
+
+            // Extract intentHash and claimant using slice
+            bytes32 intentHash = bytes32(encodedProofs[offset:offset + 32]);
+            bytes32 claimantBytes = bytes32(encodedProofs[offset + 32:offset + 64]);
+
+            // Emit event that can be proven by Polymer
+            emit IntentFulfilledFromSource(intentHash, claimantBytes, sourceChainDomainID);
```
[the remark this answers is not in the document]
i-norden: encodedProofs won't fit in a topic, but we could
1. Put it in the log data
2. Put a hash of it in a topic

The plan wasn't to support batching by having multiple intent fulfillments proven in a single event in this initial implementation which is why 1 is avoided for simplicity. 2 would work but then on the source chain the solver needs to call validate() not only with the proof returned from Prove API but also with the encodedProofs preimage.

contracts/prover/PolymerProver.sol
```diff
@@ -0,0 +1,297 @@
...
+        if (_chainIds.length != _whitelistedEmitters.length) revert SizeMismatch();
+
+        CROSS_L2_PROVER_V2 = ICrossL2ProverV2(_crossL2ProverV2);
+
+        for (uint256 i = 0; i < _chainIds.length; i++) {
+            WHITELISTED_EMITTERS[_chainIds[i]] = _whitelistedEmitters[i];
+        }
+
+        renounceOwnership();
+    }
```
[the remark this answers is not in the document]
i-norden: We want to be able to deploy this contract to the same address using CREATE2 which is dependent on constructor arguments, if we include `_whitelistedEmitters` in the constructor arguments we run into a problem because we need to know what the PolymerProver contract addresses are on all remote chains to include in the local PolymerProver constructor, but to deploy those remote PolymerProvers we would need to know what the local PolymerProver's address is to include in their construction.

contracts/prover/PolymerProver.sol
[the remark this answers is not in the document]
fish-sammy: No strong opinion. 1) is what I was thinking of, it's not even harder to do than the current implementation imo, that's not supporting batch.

contracts/prover/PolymerProver.sol
```diff
@@ -0,0 +1,297 @@
...
+     */
+    function _validateProof(
+        bytes calldata proof
+    ) internal view returns (bytes32 intentHash, address claimant, uint64 chainId) {
+        (
+            uint32 destinationChainId,
+            address emittingContract,
+            bytes memory topics,
+            /* bytes memory data */
+        ) = CROSS_L2_PROVER_V2.validateEvent(proof);
```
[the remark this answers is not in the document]
fish-sammy: Let's not do the wrapper. I am more like suggesting an interface change in CrossL2ProverV2 if it makes sense to you guys.

contracts/prover/PolymerProver.sol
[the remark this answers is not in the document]
i-norden: Roger! Sounds good to me, will support 1.

contracts/prover/PolymerProver.sol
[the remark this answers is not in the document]
i-norden: Got it! Thats good feedback, thanks! Will note this down for our next version of the contract

fish-sammy (approved):

fish-sammy (changes requested):
Blocking the merge here until it's fixed for https://github.com/eco/eco-routes/pull/330

nknavkal:
@i-norden after much back and forth on versioning, we have once again found ourselves using routes v2, but now with some new changes put in reflecting audit findings. 

here is the new polymer prover PR, if y'all could look at it and see if everything makes sense that would be great: 

https://github.com/eco/eco-routes/pull/335

PR #322, opened 2025-08-18
feat: add CREATE2_PREFIX to IntentSource to support Tron (#320)

fish-sammy:


contracts/IntentSource.sol
```diff
@@ -25,7 +25,20 @@ contract IntentSource is IIntentSource, Semver {
 
     mapping(bytes32 intentHash => VaultStorage) public vaults;
 
-    constructor() {}
+    /// @dev Tron Mainnet chain ID
+    uint256 private constant TRON_MAINNET_CHAIN_ID = 728126428;
+    /// @dev Tron Testnet (Nile) chain ID
```
nknavkal, line 30: not that it matters but this is the chainID for shasta, not nile. I think shasta is the right one to be using anyway.

PR #335, opened 2025-08-25
Polymer prover 2.0

nknavkal:
merged polymer's changes with the new changes to routes v2


contracts/prover/PolymerProver.sol
```diff
@@ -0,0 +1,238 @@
...
+        bytes32 sourceChainIdBytes32;
+        uint64 proofDataChainId;
+
+        assembly {
+            let topicsPtr := add(topics, 32)
+            let dataPtr := add(data, 32)
+
+            eventSignature := mload(topicsPtr)
+            sourceChainIdBytes32 := mload(add(topicsPtr, 32))
+            proofDataChainId := shr(192, mload(dataPtr)) // Extract first 8 bytes (64 bits) from the 32-byte word
```
i-norden, line 134: Nice that dest chainID is validated as part of the proof now. Is `IProver.challengeIntentProof` method going to be deprecated?

i-norden (approved):
LGTM

contracts/prover/PolymerProver.sol
nknavkal: most of our other routes dont have chainID-based whitelisting bc we wanted to have provers not be ownable and not have to constantly be re-deployed in the event we added a new chain, so we just had emitting contract verification, which left us vulnerable to an attack where someone deployed to the right address on a bogus chain and falsely submitted proofs from there. Afaict polymer isnt vulnerable to that, but added this just in case.

contracts/prover/PolymerProver.sol
```diff
@@ -0,0 +1,207 @@
...
+            bytes32 intentHash;
+            bytes32 claimantBytes;
+
+            assembly {
+                let dataPtr := add(decodedData, 32)
+                intentHash := mload(add(dataPtr, offset))
+                claimantBytes := mload(add(dataPtr, add(offset, 32)))
+            }
+
+            if (claimantBytes >> 160 != 0) continue;
```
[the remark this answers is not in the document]
nknavkal: @fish-sammy this has already cleared audit and is deployed, are we good to forgo this change?

PR #337, opened 2025-09-02
fix: need to abi.decode() unindexed data

notbdu:
Need to abi.decode() unindexed data to get the actual payload.


i-norden:
Fix LGTM, unit tests failed to catch this since they use a mock CrossL2ProverV2 with _data that has not been ABI encoded

RnkSngh:
> Fix LGTM, unit tests failed to catch this since they use a mock CrossL2ProverV2 with _data that has not been ABI encoded


might be worth an integration test somewhere where we generate an actual proof that is specific to the events we expect

RnkSngh (approved):

PR #340, opened 2025-09-11
fix(prover): Eco - Polymer Prover II - Spearbit audit fixes

i-norden:


i-norden:
Completely agree inre the whitelist. Last fix to the MAX_LOG_DATA_SIZE in with the last commit 👍

PR #372, opened 2026-03-04
fix: add per-address nonce to route salt to prevent same-block hash c…

nknavkal:
…ollisions

Without extra entropy, two createIntent() calls in the same block with the same amount on the same deposit address produce an identical route salt, reward deadline, and therefore identical intentHash. When the second call hits onlyFundable the modifier short-circuits silently, leaving tokens stranded in the deposit address with a dangling approval.

Add a private _nonce counter to BaseDepositAddress, incremented on every createIntent() call before _executeIntent runs. Each derived contract mixes _currentNonce() into its route salt, guaranteeing a distinct intentHash regardless of block.timestamp or amount.

Tests added to all three DepositAddress variants to verify same-block, same-amount calls produce distinct intent hashes.

## Pull Request Title Format

Please ensure your PR title follows the [Conventional Commit](https://www.conventionalcommits.org/) format:

```
<type>[optional scope]: <description>
```

### Allowed Types:

- `feat`: A new feature
- `fix`: A bug fix
- `docs`: Documentation only changes
- `style`: Changes that do not affect the meaning of the code
- `refactor`: A code change that neither fixes a bug nor adds a feature
- `test`: Adding missing tests or correcting existing tests
- `chore`: Changes to build process or auxiliary tools
- `perf`: A code change that improves performance
- `ci`: Changes to CI configuration files and scripts
- `build`: Changes that affect the build system or external dependencies
- `revert`: Reverts a previous commit

### Examples:

✅ `feat: add cross-chain intent validation`
✅ `fix: resolve memory leak in prove method`
✅ `docs: update API documentation for routes`
✅ `refactor(inbox): simplify refund logic`
✅ `test: add edge case coverage for token transfers`
✅ `chore: update dependencies to latest versions`

### Important Notes:

- The description should be lowercase and not end with a period
- Use present tense ("add" not "added" or "adds")
- Scopes are optional but helpful for organizing changes
- If you need to bypass this check temporarily, add the label `skip-pr-title-check`

---

## Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Test improvements
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] CI/Build changes


contracts/deposit/BaseDepositAddress.sol
```diff
@@ -72,6 +76,9 @@ abstract contract BaseDepositAddress is ReentrancyGuard {
         uint256 amount = IERC20(sourceToken).balanceOf(address(this));
         if (amount == 0) revert ZeroAmount();
 
+        // Increment nonce before execution so derived contracts read the updated value
+        ++_nonce;
+
         // Execute variant-specific intent creation
         intentHash = _executeIntent(amount);
```
fish-sammy: Why not just pass `++_nonce` into `_executeIntent` so that you don't even need the `_currentNonce` function and also avoids extra storage read?

nknavkal: nice

fish-sammy (approved):

PR #381, opened 2026-04-21
feat: add ERC20 gateway deposit variant and approval-based intent creation

carlosfebres:
## Summary

Extends the CCTP-based deposit address system along two complementary axes:

1. **ERC20 variant for non-Arc destinations.** A new deposit address / factory pair targeting chains where USDC arrives as a standard ERC20 (Base, Optimism, Arbitrum, Polygon, …) instead of as the native gas token (Arc).
2. **Approval-based intent creation.** A new entry point on `BaseDepositAddress` that lets users sign a standard ERC20 approval to the deterministic deposit address, removing the need for a trusted operator wallet to custody funds before the intent is triggered.

Both changes are inherited by every deposit variant, so the existing Arc flow also picks up the approval entry point.

### ERC20 gateway deposits
The existing `DepositFactory_CCTPMint_Gateway` was shaped around Arc, where USDC is the native gas token. On every other EVM chain, the destination-side intent has to move USDC as a regular ERC20 — a different token structure with no native-amount scaling. Rather than stretching the Arc contract with conditional paths, this branch introduces a dedicated variant with destination-agnostic naming so the same contract can be reused across EVM chains. It's been deployed at the same CREATE3 address on Base, Optimism, and Arbitrum.

### Approval-based flow
Deposit flows today lean on an operator wallet to receive funds and trigger the intent. That wallet is mutable, human-controlled, and has to be trusted by the user. With the new flow, the user signs a plain ERC20 approval to the **deposit address itself** — a deterministic, CREATE2-derived contract whose address can be computed off-chain before it's even deployed. The contract then pulls the lesser of allowance and balance when the intent is created. Net effect: no custodial operator in the path, and the UX reduces to a single approve that anyone can later crank.

## Type of Change

- [x] New feature (non-breaking change which adds functionality)

## Scope

- 1 new deposit variant (`DepositAddress_CCTPMint_GatewayERC20` + `DepositFactory_CCTPMint_GatewayERC20`)
- 1 cross-cutting addition to `BaseDepositAddress` that all variants inherit
- CREATE3 deployment script for same-address multi-chain deployment
- Forge test coverage for both additions (existing test style preserved)

## Test plan

- [ ] `forge test` — full suite green
- [x] Targeted runs: `DepositAddress_CCTPMint_GatewayERC20.t.sol`, `DepositFactory_CCTPMint_GatewayERC20.t.sol`, and the new `createIntentWithApproval` cases in `BaseDepositAddress`
- [x] Sanity check: deployed addresses on Base / Optimism / Arbitrum match the CREATE3 prediction for `0x5947DfA1Ef80Fc4c6c6aECE2dF350429162cbEFb`
- [x] Regression: existing Arc-variant behavior unchanged end-to-end

### DIFF
--- a/contracts/deposit/DepositAddress_CCTPMint_GatewayERC20.sol
+++ b/contracts/deposit/DepositAddress_CCTPMint_GatewayERC20.sol
@@ -1,0 +1,313 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.26;
+
+import {IERC20} from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import {BaseDepositAddress} from "./BaseDepositAddress.sol";
+import {Portal} from "../Portal.sol";
+import {Intent, Route, Reward, TokenAmount, Call} from "../types/Intent.sol";
+import {DepositFactory_CCTPMint_GatewayERC20 as DepositFactory} from "./DepositFactory_CCTPMint_GatewayERC20.sol";
+
+/**
+ * @title DepositAddress_CCTPMint_GatewayERC20
+ * @notice Minimal proxy contract that constructs two Intent structs for CCTP+Gateway transfers to ERC20 destinations
+ * @dev Creates TWO intents to bridge USDC from source chain to a destination via CCTP, then deposit into Gateway:
+ *      Intent 2 (published first): Gateway deposit on destination — receives CCTP-bridged USDC and deposits into Gateway
+ *      Intent 1 (published and funded second): CCTP burn on source chain — burns USDC and mints to Intent 2's vault
+ *      Each DepositAddress is specific to one user's destination address.
+ *      Deployed via CREATE2 by DepositFactory_CCTPMint_GatewayERC20 for deterministic addressing.
+ *      Unlike the Arc variant, USDC is a standard ERC20 on the destination (no decimal scaling needed).
+ *
+ * @dev Intent Call Flow:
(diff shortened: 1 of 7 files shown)


nknavkal (approved):
lgtm. 
- can have salt collisions if two intents are created in same block. This results in one intetnt's funds remaining in the DA, but they can be rescued easily via a call to createIntent, so if this is deployed already we dont need to patch

contracts/deposit/DepositAddress_CCTPMint_GatewayERC20.sol
```diff
@@ -0,0 +1,385 @@
...
+            destinationChainId,
+            portalAddress,
+            destinationProverAddress,
+            destinationUsdc,
+            gatewayAddress,
+            netAmount,
+            salt,
+            deadline
+        );
+        (, vault2) = portalContract.publish(intent2);
```
fish-sammy, line 204: You shouldn't have to publish it. You can call `.intentVaultAddress` instead. I imagine publishing intent2 on the source chain is confusing to the solver.

This function exists on the Portal contract.
```
    function intentVaultAddress(
        uint64 destination,
        bytes memory route,
        Reward calldata reward
    ) external view returns (address);
```

fish-sammy (approved):

contracts/deposit/DepositAddress_CCTPMint_GatewayERC20.sol
carlosfebres: Yeah, the publication is intentional 👍🏻

PR #382, opened 2026-04-27
feat: layerzero permissionlessness

nknavkal:
**Summary**

Builds on #379 (Tron support). This PR adds two things on top of that baseline:

**revokeDelegation() on LayerZeroProver**

Adds a `revokeDelegation()` function callable only by the current LZ delegate. It sets the delegate to `address(this)`, permanently making all privileged LZ endpoint actions (`setConfig`, `setSendLibrary`, `skip`, etc.) uncallable — since the contract itself cannot initiate transactions. This is a one-way operation that makes the prover's LZ configuration immutable once invoked, enabling permissionless operation.

**Script improvements**

- `configure-lz-evm-tron.ts` — refactored to accept chain IDs as CLI arguments instead of a single hardcoded chain; EVM and Tron LZ endpoints now auto-resolved from `lzDeployments.json` rather than hardcoded; mainnet DVN policy updated to 1 required (LayerZero Labs) + 1-of-2 optional (Nethermind, Deutsche Telekom)
- `deploy-evm-tron.ts` — deploy and configure steps parallelized across chains

**Type of Change**

- New feature (revokeDelegation for permissionless LZ config)
- Code refactoring (script improvements)


contracts/prover/LayerZeroProver.sol
```diff
@@ -256,13 +292,29 @@ contract LayerZeroProver is ILayerZeroReceiver, MessageBridgeProver, Semver {
...
-                unpacked.gasLimit // gas amount
-            );
+        // Compute a gas floor proportional to the number of intents in the batch.
+        // Each intent requires a cold SSTORE + event emission in _processIntentProofs.
+        // The 8-byte header is subtracted before dividing by 64 (bytes per intent pair).
+        uint256 numIntents = encodedProofs.length > 8
+            ? (encodedProofs.length - 8) / 64
+            : 0;
+        // gasFloor is bounded by message size and the fixed constants — never overflows uint128.
+        uint128 gasFloor = uint128(MIN_GAS_LIMIT + numIntents * GAS_PER_INTENT);
```
fish-sammy: Use `SafeCast` instead.

contracts/IntentSourceTron.sol
```diff
@@ -0,0 +1,892 @@
...
+import {VaultTron} from "./vault/VaultTron.sol";
+import {Clones} from "./vault/Clones.sol";
+
+/**
+ * @title IntentSourceTron
+ * @notice IntentSource variant for Tron chains. Identical to IntentSource except it
+ *         deploys VaultTron clones instead of Vault clones, enabling support for
+ *         non-standard ERC20 tokens such as Tron USDT.
+ */
+abstract contract IntentSourceTron is OriginSettler, IIntentSource {
```
fish-sammy: I don't like this being mostly a copy of IntentSource instead of picking up the same inheritance pattern used for the Vault and LocalProver. It should be achivable by changing the constructor of the IntentSource to take two parameters address vaultImplementation and bytes1 create2Prefix. 

1. Regular Portal's constructor does `IntentSource(address(new Vault), 0xff)`
2. Tron Portal's constructor does `IntentSource(address(new VaultTron()), 0x41)`

fish-sammy (changes requested):

contracts/IntentSourceTron.sol
```diff
@@ -0,0 +1,892 @@
...
+            return true;
+        }
+
+        uint256 remaining = rewardAmount - balance;
+        uint256 transferAmount = remaining
+            .min(token.allowance(funder, address(this)))
+            .min(token.balanceOf(funder));
+
+        if (transferAmount > 0) {
+            token.safeTransferFrom(funder, vault, transferAmount);
```
fish-sammy: Is this not a problem for Tron USDT? If not, add a comment so that future readers won't see it as a bug.

fish-sammy (dismissed):

contracts/IntentSource.sol
```diff
@@ -40,10 +39,12 @@ abstract contract IntentSource is OriginSettler, IIntentSource {
...
-     * @dev Deploys vault implementation
+     * @param vaultImplementation Address of the vault implementation used for cloning
+     * @param create2Prefix CREATE2 prefix byte for the target chain (0xff for EVM, 0x41 for Tron)
      */
-    constructor() {
-        VAULT_IMPLEMENTATION = address(new Vault());
+    constructor(address vaultImplementation, bytes1 create2Prefix) {
+        VAULT_IMPLEMENTATION = vaultImplementation;
+        CREATE2_PREFIX = create2Prefix;
     }
```
hjorthjort, line 48: Could you not use the chainID instead of manually passing parameters (which seems more error-prone)?

hjorthjort: Actually seems this was the case before this PR: https://github.com/eco/eco-routes/pull/382/changes#diff-829e18233f2ced897c50ff96e8ba8b7a9bc826e85f5e0d5af8d6b42b2f8f6316L51-L59

fish-sammy (approved):

PR #409, opened 2026-07-07
feat: per-deployment native/ERC20 alias support in recovery + reward validation

re1ro:
## Summary
- Adds an optional immutable `NATIVE_ERC20` address to `IntentSource` (threaded through the `Portal`/`PortalTron` constructors as a new `nativeErc20` param). Default deploy value is `address(0)`, which leaves every check below fully inert and preserves existing behavior on every current deployment.
- When configured, a deployment's native asset may be aliased to an ERC20 token (its balance mirrors `address(this).balance` on every vault). Two things follow from treating them as the same underlying funds rather than independent assets:
  - `recoverToken` now also blocks recovery of the alias token whenever the reward still carries a non-zero native amount.
  - A reward may not declare both a non-zero native amount and an explicit token leg for the alias address at the same time — rejected up front via the existing `RewardTokensNotUnique` error, on every entry point that can escrow a reward (`publish`, `fund`, `fundFor` — the latter previously had no reward-leg validation on any entry point).
- Updates every `Portal`/`PortalTron` construction call site for the new constructor parameter, and adds regression tests covering the default inert case and both alias-configured guard paths (including a direct `fundFor` call with no prior `publish`, to confirm the guard isn't bypassable through that entry point).

Supersedes #408, which got closed as a side effect of a branch rename rather than a merge.

## Test plan
- [x] `forge build`
- [x] `forge test` (573 passing)
- [x] `yarn test:hardhat` (112 passing)
- [x] `yarn test:ts` (42 passing)
- [x] `forge build --sizes` — Portal/PortalTron runtime margin 1,554 bytes under EIP-170

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01K4pnnevHt2zFJwDqPyJreT

### DIFF
--- a/contracts/IntentSource.sol
+++ b/contracts/IntentSource.sol
@@ -258,6 +258,8 @@ abstract contract IntentSource is OriginSettler, IIntentSource {
         bytes memory route,
         Reward memory reward
     ) public returns (bytes32 intentHash, address vault) {
+        _requireNoNativeAliasConflict(reward);
+
         (intentHash, , ) = getIntentHash(destination, route, reward);
         vault = _getVault(intentHash);
 
@@ -628,6 +630,10 @@ abstract contract IntentSource is OriginSettler, IIntentSource {
         address funder,
         bool allowPartial
     ) internal onlyFundable(intentHash) {
+        // Reject a reward-token leg regardless of entry point. `publish` already checks this, but
+        // `fund`/`fundFor` can escrow an intent directly without a prior `publish` call.
+        _requireNoNativeAliasConflict(reward);
+
         bool fullyFunded = _fundNative(vault, reward.nativeAmount);
 
         uint256 rewardsLength = reward.tokens.length;
@@ -723,6 +729,10 @@ abstract contract IntentSource is OriginSettler, IIntentSource {
         address funder,
         address permitContract
     ) internal onlyFundable(intentHash) returns (address vault) {
+        // Reject a reward-token leg regardless of entry point — see the matching check and comment in
+        // {_fundIntent}.
+        _requireNoNativeAliasConflict(reward);
+
         vault = _getOrDeployVault(intentHash);
         bool fullyFunded = IVault(vault).fundFor{value: msg.value}(
             reward,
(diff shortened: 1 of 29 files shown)


contracts/IntentSource.sol
```diff
@@ -843,6 +894,14 @@ abstract contract IntentSource is OriginSettler, IIntentSource {
                 revert InvalidRecoverToken(token);
             }
         }
+
+        if (
+            NATIVE_ERC20 != address(0) &&
```
Foivos: This line is unnecessary since `token != address(0)` (above) && `token == NATIVE_ERC20` (next line) imply that.

Foivos (approved):

Foivos (approved):

Foivos (approved):

MSevey:
## Testing gap analysis

The guard logic itself looks correct, and the double-count/blocking paths are well covered. The gaps below are about **test coverage**, not the implementation. Filing this so the audit follow-up (PAR-369) has something concrete, and to answer Carlos's "we need unit tests that simulate this duality — how?" from the thread.

Framing the reward as three shapes (native and the alias ERC-20 are two interfaces onto one balance):

| Reward shape | Escrow (publish/fund) | Withdraw payout | Refund payout | Recover |
|---|---|---|---|---|
| **native + alias leg** → blocked | ✅ all 3 entry points | n/a | n/a | ✅ blocked |
| **native-only** → allowed | ⚠️ setup only, no assertion | ❌ untested | ❌ untested | ✅ (unrelated-token case) |
| **alias-ERC20-leg-only** → allowed | ❌ untested | ❌ untested | ❌ untested | ❌ untested |

### Are the valid shapes tested? Mostly no.

- **native + alias leg (blocked)** is the well-covered case: `publish`, `fund`, `fundFor` reverts + the recover block. 👍
- **native-only (allowed)** is only exercised as *setup* for the recover tests — nothing asserts a native-only USDC reward actually withdraws to the claimant or refunds correctly on an alias deployment.
- **alias-ERC20-leg-only (allowed)** has no coverage at all. Heads up: `testRecover_aliasToken_allowedWhenNoNativeLeg` looks like it covers this, but it uses an **empty** reward (no tokens, no native) and recovers a stray mint. A real reward carrying the alias token as an ERC-20 leg (`nativeAmount=0, tokens=[USDC]`) is never published/funded/withdrawn/refunded, and the "recover of that leg is blocked by the existing `reward.tokens` loop" branch is untested.

### Gaps to close, prioritized

**P1 — the duality is never actually simulated.** Every test uses a plain `TestERC20` whose ERC-20 balance does *not* mirror native. So the tests prove the guard *reverts on declaration*, but never that the guard closes the underlying same-balance drain path on an Arc-like token. This is the same point Carlos raised, and it's the one that matters for the audit. Proposed: a `MockDualInterfaceToken` (ERC-20 `transfer`/`transferFrom` also moves native, and vice versa — Arc semantics), then a regression test showing the recover/refund drain path is open before the fix and closed after.

**P1 — alias-ERC20-leg-only (row 2) has no coverage.** This is a first-class valid shape on Arc. Add fund/withdraw/refund + recover-blocked tests for `nativeAmount=0, tokens=[alias]`.

**P2 — happy-path payouts on an alias deployment.** `Vault.t.sol` proves native withdraw/refund only on the default portal. Add native-only reward → withdraw pays claimant, → refund returns to creator, on an alias-configured portal (ideally the dual-interface mock).

**P2 — ERC-7683 entry points not exercised for the guard.** Covered via `publish`/`fund`/`fundFor`, but not via `publishAndFund`, `publishAndFundFor`, or `open`/`openFor`. These route through `_fundIntent`/`_fundIntentFor` so they should revert, but the gasless `openFor` path is exactly the untrusted-input surface worth a direct test.

**P3 — alias token at a non-zero index:** every revert test puts the alias as the sole leg; add `tokens=[realToken, aliasUSDC]` + native to confirm the loop scans all legs.

**P3 — partial funding (`allowPartial=true`)** on an alias deployment (funding one interface bumps the other's `balanceOf`).

**P3 — `PortalTron`** got the same constructor change but has no non-zero-alias test (academic since Tron native is TRX, but the path is untested).

**Not a gap, for the record:** default-deployment inertness is structurally guaranteed — both `_requireNoNativeAliasConflict` and `_validateRecover` short-circuit on `NATIVE_ERC20 == address(0)`, and the existing suite runs against `address(0)` portals. A "default still allows native+token" test would document intent cheaply but isn't strictly required.

### Suggested single test file

A `MockDualInterfaceToken` + tests covering: (1) exploit regression on the dual-balance token, (2) native-only lifecycle, (3) alias-ERC20-only lifecycle, (4) ERC-7683 `open`/`openFor` revert, (5) multi-leg + partial funding. That closes P1/P2 and gives PAR-369 concrete evidence.

*(Analysis generated with Claude Code, reviewed by @MSevey.)*

MSevey (approved):

PR #410, opened 2026-07-08
feat: adopt solver-style semantic-release with contract version rewriting

carlosfebres:
## Summary

Replaces the dead beta-branch, deploy-coupled release pipeline with a solver-v2-style semantic-release flow on `main`, and cleans out the dead deployment machinery. Closes PAR-372.

- **New release flow** (`.github/workflows/release.yaml` + `.releaserc.json`): on push to `main`, semantic-release computes the next version from conventional commits, rewrites every contract `version()` function to the full `x.y.z` (new `scripts/release/update-versions.ts`, 5 Jest tests), bumps `package.json`, updates `CHANGELOG.md`, commits back as `chore(release): x.y.z [skip ci]`, then tags and publishes a GitHub release. **No contract deploys, no npm publish.**
- **PR title enforcement** (`.github/workflows/pr-title-check.yml`): solver's SHA-pinned `amannn/action-semantic-pull-request` config, scoped to `main`.
- **Dead pipeline removed**: `trigger-from-chains.yaml`, `scripts/semantic-release/` (27 files), `scripts/utils/`, `deployRoutes.sh` / `verifyRoutes.sh` / `deploySingletonFactory.sh`, `scripts/createXOld/`, 13 dead package.json scripts (many pointed at files that no longer exist), 4 unused devDependencies.
- **Orphans removed**: `executeInstructions.ts` (printed generated private keys to console), `package_working.json` (stale pre-fork ecoism package.json).
- **Docs**: `scripts/README.md` rewritten; `README.md`/`CLAUDE.md` release+deploy sections updated, including fixing the long-broken `forge script script/Deploy.s.sol` path (the file lives at `scripts/Deploy.s.sol`); two design docs moved from repo root into `CLAUDE/specs/`.

Design spec: `CLAUDE/specs/2026-07-08-semver-release-design.md` · Plan: `CLAUDE/plans/2026-07-08-semver-release-alignment.md`

## ⚠️ Prerequisites before the first release (human actions)

1. **Tag pruning + version baseline** (per review): the latest actually-released version is `v2.8.17`; tags above it (`v3.x`, `v9.x`, prereleases) were never released. Admin: prune tags newer than `v2.8.17`, then baseline main with `git tag v2.8.18 <main tip> && git push origin v2.8.18`. `v3.0.0` is cut later via a `BREAKING CHANGE` commit. This is now **CI-enforced**: the release job aborts unless a stable tag >= `v2.8.17` is reachable from `main`.
2. **Branch protection**: `main` must allow the release workflow's push-back commit (add `github-actions[bot]` to the bypass list, or use an app/PAT token).
3. **Merge method**: verify the repo uses squash-merge for PRs (the PR-title check assumes squash commits are what semantic-release analyzes).
4. **Retire old machinery** (optional, after merge): the `beta` branch and old-pipeline-only secrets (AWS role, `VERIFICATION_KEYS`, `RELEASE_PAT`, npm token in the release context).

## Notes

- The version string is compiled into bytecode, so every release changes contract bytecode and therefore CREATE2 deterministic addresses — full `major.minor.patch` granularity was an explicit design decision.
- No files under `contracts/` change in this PR (verified: `git diff origin/main..HEAD -- contracts/` is empty). Version rewriting happens at release time only.
- First release's commit will also sweep ~15 pre-existing format-drifted contracts via the prettier pass in prepareCmd (formatting only; `bytecode_hash = "none"` so artifacts are unaffected).
- Pre-existing issues surfaced but not addressed here: `yarn lint` exits 1 from repo-wide format drift on main; depcheck reports missing dep `borsh` used by `test/deposit/BorshEncodingVerification.test.ts`.

## Test plan

- [x] `yarn test` full suite: forge 567/567, hardhat 112, jest 5/5 (684 total, 0 failures)
- [x] semantic-release dry run with temporary local baseline tag: picks up the baseline, computes the next version correctly, zero npm steps
- [x] Baseline guard verified both ways: FAILs on current main (v1.6.1 max), PASSes with a baseline tag
- [x] Version tests green against source rewritten to 9.9.9 (the post-release state that previously failed)
- [x] Release rehearsal: `update-versions.ts 3.9.9` + prettier + `forge build` on rewritten sources succeeds; rewrite touches exactly `Semver.sol`, `TestProver.sol`, `TestMessageBridgeProver.sol` + `package.json`; clean revert
- [x] Invalid CLI input exits non-zero before any file writes (aborts the release)
- [x] Repo-wide grep: zero references to any deleted file
- [ ] After merge: admin prunes stale tags + creates `v2.8.18` baseline tag, then verify the first real release on a trivial `fix:`/`docs:` commit


.releaserc.json
```diff
@@ -1,64 +1,28 @@
...
-      "channel": "beta"
-    }
-  ],
-  "repositoryUrl": "git@github.com:eco/eco-routes.git",
-  "lastRelease": {
-    "version": "2.1.5",
-    "gitTag": "v2.1.5",
-    "gitHead": "1c1fdd412f40a7d2e13ac357173da73fe1416559"
-  },
+  "branches": ["main"],
```
MSevey, line 2: 🔴 **P1 — the release baseline tag is unreachable from `main`.** The highest tag that's an ancestor of `main` is `v1.6.1` (verified via `git merge-base --is-ancestor`); `v3.2.6` / `v9.x` / `v3.3.0-alpha.1` live on other branches and aren't reachable. Releasing off `main` with no baseline tag makes semantic-release compute from the v1.x line, and `scripts/release/update-versions.ts` then **downgrades** `package.json` and `Semver.sol` into that range — regressing the deployed contract's `version()` (and its CREATE2 address) below the real v3/v9 history. The PR body documents the `git tag v3.3.0` prereq, but nothing in CI enforces it, so a merge-then-forget window ships a regressed release. Consider a guard step in `release.yaml` that aborts if no `v3.x` tag is reachable.

MSevey, line 2: 💡 **Versioning strategy for majors.** Because releases bump off `main`, once a major lands there's no branch to patch the previous major in prod. When `main` bumps `v2.x → v3.x`, cut a `release-v2.x` branch from the last `v2.x` tag (`v2.8.17`) so hotfixes can be cherry-picked there while `main` stays on the latest line. Worth adding to the release design doc alongside the branches config.

.releaserc.json
```diff
@@ -1,64 +1,28 @@
...
         }
       }
     ],
     "@semantic-release/release-notes-generator",
     "@semantic-release/changelog",
-    "./dist/semantic-release/index.js",
+    [
+      "@semantic-release/exec",
+      {
+        "prepareCmd": "npx tsx scripts/release/update-versions.ts ${nextRelease.version} && npx prettier --write 'contracts/**/*.sol'",
```
MSevey: 🟡 `prettier --write 'contracts/**/*.sol'` sweeps every format-drifted contract into the `chore(release)` commit — 15 files are currently non-compliant (`CCIPProver.sol`, `HyperProver.sol`, `LayerZeroProver.sol`, `vault/Vault.sol`, the `deposit/*` set, …). Cosmetic for bytecode (`bytecode_hash="none"`), but it pollutes deployed-source history and is exposed to `prettier-plugin-solidity` version drift between local and CI. Cleaner to land a format-only PR first so release commits are version-only. Separately, there's no `forge build` before this commits — a bad rewrite would ship uncompiled.

scripts/release/update-versions.ts
```diff
@@ -0,0 +1,82 @@
...
+ * Usage: npx tsx scripts/release/update-versions.ts <x.y.z>
+ *
+ * The version string is compiled into bytecode; changing it changes CREATE2
+ * deployment addresses. Full major.minor.patch is intentional (see
+ * CLAUDE/specs/2026-07-08-semver-release-design.md).
+ */
+import * as fs from 'fs'
+import * as path from 'path'
+
+const VERSION_FUNCTION_REGEX =
```
MSevey, line 16: 🟡 Correct for today's one-line bodies (matches exactly `Semver.sol` + the two test provers and correctly skips the `ISemver` interface declaration). Two latent risks worth a guard: `\{[^}]*\}` truncates at the first `}`, so any future `version()` whose body contains a nested block would be rewritten into non-compiling source; and a body with different whitespace (e.g. a newline before `{`) is silently skipped — and this runs *before* prettier, so it sees pre-formatted source. A hard assert on the expected number of rewritten files would turn a bad/incomplete rewrite into a loud failure instead of a silent one.

.github/workflows/release.yaml
```diff
@@ -1,125 +1,37 @@
...
-          git checkout beta
-          echo "Current branch: $(git rev-parse --abbrev-ref HEAD)"
+        run: yarn install --frozen-lockfile
 
-      # Run semantic-release with access to the AWS secrets
-      - name: Run Semantic Release
-        id: semantic-release
+      - name: Run semantic-release
+        id: semantic
+        run: npx semantic-release
```
MSevey, line 56: 🟡 No build/test gate before release. This job runs only `yarn install` then `npx semantic-release` — the regex-rewritten Solidity is committed and tagged with no `forge build` / `forge test`. A broken `main` (or a bad rewrite from the regex) would be released unverified. Consider gating the semantic-release step behind a `forge build` (or `yarn test`).

.github/workflows/release.yaml
```diff
@@ -1,125 +1,37 @@
...
     outputs:
-      new_version: ${{ steps.semantic-release.outputs.new_version }}
-      release_notes: ${{ steps.semantic-release.outputs.release_notes }}
-    permissions:
-      contents: write # Needed for github release
-      issues: write # Needed for issue comments
-      pull-requests: write # Needed for PR comments
-      id-token: write # Needed for npm publishing with provenance
-
+      new_release_published: ${{ steps.semantic.outputs.new_release_published }}
```
MSevey, line 17: ℹ️ These job outputs are dead wiring — the `successCmd` in `.releaserc.json` does correctly populate `$GITHUB_OUTPUT`, but no downstream job or workflow (`needs:`) consumes `new_release_*`. Harmless; either drop them or wire a consumer. (Confirmed there's no infinite release loop — `GITHUB_TOKEN`-authored pushes don't retrigger workflows, so the `[skip ci]` is just belt-and-suspenders.)

MSevey (commented):
## `/review` findings — semantic-release adoption

Reviewed with Claude (structured + adversarial) and Codex (gpt-5.5, high). Most findings are inline below. Two items live here because they can't anchor inline — the first touches test files outside this diff, and the last two are process prereqs.

### 🔴 P1 — First release turns `main` red (not addressed anywhere)

`contracts/libs/Semver.sol` returns `"2.6"` today, and three tests assert that literal:
- `test/libs/Semver.t.sol:23` — `assertEq(version, "2.6")`
- `test/core/Inbox.t.sol:47` — `assertEq(portal.version(), "2.6")` (Portal inherits Semver)
- `test/prover/CCIPProver.t.sol:130` — `assertEq(ccipProver.version(), "2.6")` (inherits Semver)

The first release rewrites `version()` to the full `x.y.z`. The release commit is `[skip ci]`, but `ci.yaml` runs `forge test` on every push, so the **next** push to `main` fails these three assertions and leaves `main` red until someone edits them. This happens even with a correct `3.3.0 → 3.4.0` bump — it's independent of the baseline-tag prereq. Fix: make these assertions version-agnostic (assert a semver shape) rather than a literal.

### 🧹 Prereq before merge — prune stale git tags

The tag history is noisy. Before enabling releases off `main`, delete all the tags "newer" than the latest stable **`v2.8.17`**.  For example we have lots of `v3.x` tags that definitely have not been released and we are working on v3 now so those should be dropped so we can just start fresh with v3.0.0 when we are ready. A cleaner tag set removes ambiguity from the baseline semantic-release computes and makes the go-live tagging step unambiguous.

### 💡 Versioning strategy — cut a release branch when bumping majors

Because versions bump off `main`, once a major lands there's no branch to patch the previous major in production. Recommend: when `main` bumps `v2.x → v3.x`, create a `release-v2.x` branch from the last `v2.x` tag (`v2.8.17`) so production fixes can be cherry-picked there while `main` carries the latest line. (Also noted inline on the `branches` config.)

Everything else — baseline landmine, prettier sweep, regex fragility, missing build gate, dead outputs — is inline.

.releaserc.json
carlosfebres: Guard added in cd1e6a17: the release job now aborts unless a stable tag >= v2.8.17 is reachable from the branch ("Verify release baseline tag" step), so the merge-then-forget window fails loudly instead of shipping a v1.x-based downgrade. Verified both directions locally: current main state → guard FAILs; with a baseline tag → PASSes. Baseline plan updated per your top-level comment: prune tags > v2.8.17, tag main v2.8.18, v3.0.0 via BREAKING CHANGE commit (docs in scripts/README.md + CLAUDE.md, 9c3e0e85).

carlosfebres: Documented in 9c3e0e85: scripts/README.md gains a "Patching a previous major" section (cut release-vN.x from the previous major's last tag on any major bump) and the design doc's amendment section records it.

.releaserc.json
carlosfebres: Both addressed: (1) prepare now formats only the files it rewrote (prettier invoked from update-versions.ts on the updated list, 657c6512) so release commits stay version-only regardless of repo drift — no dependency on a format-only PR landing first; (2) prepareCmd now runs forge build on the rewritten source before the git plugin commits/tags, and the job runs forge test before semantic-release at all (cd1e6a17).

scripts/release/update-versions.ts
carlosfebres: Fail-loud guard added (657c6512): the script now throws (exit 1) if zero files were rewritten, so a whitespace-drifted or restructured version() aborts the release instead of silently tagging unchanged source. Went with a zero-check rather than an exact-count assert so adding a new ISemver contract doesn't require touching the release script; the nested-block truncation case is covered by the new forge build in prepareCmd — truncated output doesn't compile, so it fails before commit/tag. Also: prettier (house multi-line format) output is round-trip tested — the regex matches both single-line and formatted forms.

.github/workflows/release.yaml
carlosfebres: Added in cd1e6a17: foundry-toolchain install + forge test gate before semantic-release, and forge build inside prepareCmd so the rewritten source must compile before the release commit exists.

.github/workflows/release.yaml
carlosfebres: Keeping them: they're spec'd as the interface for future downstream jobs (helm-style publish in solver consumes the same trio), and successCmd already populates them. Wiring a consumer now would be speculative; dropping them re-adds churn when a consumer lands. Noted on the [skip ci] confirmation.

scripts/release/tests/update-versions.spec.ts
```diff
@@ -0,0 +1,101 @@
...
+
+describe('update-versions', () => {
+  let rootDir: string
+
+  beforeEach(() => {
+    rootDir = fs.mkdtempSync(path.join(os.tmpdir(), 'update-versions-'))
+    fs.mkdirSync(path.join(rootDir, 'contracts', 'libs'), { recursive: true })
+    fs.writeFileSync(
+      path.join(rootDir, 'contracts', 'libs', 'Semver.sol'),
+      VERSIONED_CONTRACT,
```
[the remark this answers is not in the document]
carlosfebres: Test added in 657c6512: bumps an already-rewritten single-line version() from 3.2.7 to 3.2.8 and asserts exactly one file updated with the new literal (plus a zero-match test covering the new fail-loud path). 7/7 passing.

carlosfebres:
## Review findings addressed — all validated, all fixed

**P1 (main turns red):** confirmed — three tests asserted the literal `"2.6"` and would have failed the first post-release push. Fixed in cfadc7ee: assertions now check semver shape via a shared `BaseTest._assertValidSemver` helper. Verified by running the affected suites against source rewritten to `9.9.9` — green both before and after rewrite.

**Tag pruning:** adopted. Docs now say: prune everything newer than `v2.8.17` (all `v3.x`, the stray `v9.0.0`/`v9.0.1`, prerelease tags), baseline main at `v2.8.18`, and cut `v3.0.0` later via a `BREAKING CHANGE` commit (9c3e0e85). Note `v2.8.17` itself is on the `beta` lineage and not reachable from `main` either — hence the explicit baseline tag, and it's now **CI-enforced**: the release job aborts unless a stable tag ≥ `v2.8.17` is reachable (cd1e6a17).

**Release-branch-for-majors:** documented in `scripts/README.md` + the design doc amendment.

**Inline items** (each answered in-thread): forge test gate + forge build on rewritten source before commit/tag; prepare fails loudly on zero rewrites; prettier scoped to only the rewritten files (release commits stay version-only — no format-PR dependency); bump-path test added; job outputs kept deliberately (spec'd interface for future downstream jobs).

New commits: cfadc7ee, 657c6512, cd1e6a17, 9c3e0e85. Full suite green: forge 567/567, jest 7/7.

carlosfebres:
**Tag pruning done** (Carlos confirmed the v3.x line was pre-release). All 24 tags newer than `v2.8.17` deleted from origin + local: `v2.9.0-alpha.1`, `v3.0.0-alpha.1`–`.14`, `v3.2.0`, `v3.2.3`–`v3.2.6`, `v3.3.0-alpha.1`, `v9.0.0`, `v9.0.1`, and the malformed `vv2.0.1510`. `v2.8.17` is now the highest tag in the repo.

Remaining go-live step from the prereqs: baseline tag `v2.8.18` on main's tip after this merges (the release workflow's guard enforces it).

MSevey (commented):
Re-review of the fixes: the baseline guard, prettier scoping, forge build/test gate, semver-shape test assertions, and the tag-pruning/branch-strategy docs all validate correctly (guard shell logic independently confirmed: fails on empty / v1.x / v2.8.16, passes on v2.8.17+). Two residual items below.

scripts/release/update-versions.ts
```diff
@@ -0,0 +1,94 @@
...
+export function main(argv: string[]): void {
+  const version = argv[2]
+  if (!version || !SEMVER_REGEX.test(version)) {
+    throw new Error(
+      `Usage: npx tsx scripts/release/update-versions.ts <x.y.z> (got: ${version ?? 'nothing'})`,
+    )
+  }
+  const rootDir = process.cwd()
+  const updated = updateSolidityVersions(rootDir, version)
+  if (updated.length === 0) {
```
MSevey, line 89: 🟡 The fail-loud guard is a real improvement, but it doesn't pin the **production** file. This regex matches 3 files — `contracts/libs/Semver.sol` (deployed; Portal inherits it) plus `contracts/test/TestProver.sol` and `TestMessageBridgeProver.sol`. If `Semver.sol` alone drifts out of the regex shape (a modifier added to `version()`, or a non-standard reformat) while the two test helpers still match, `updated.length` is `2` → this check passes → `forge build` passes (the test provers compile fine) → the release ships with the **deployed `Semver` version stale**. That's exactly the silent-stale-version case this guard is meant to prevent, just narrowed to the production file.

One-line fix that keeps your zero-check rationale (no exact-count brittleness): also assert the canonical file was rewritten, e.g.
```ts
if (!updated.some((f) => f.endsWith(path.join('contracts', 'libs', 'Semver.sol')))) {
  throw new Error('contracts/libs/Semver.sol was not rewritten — deployed version would be stale. Aborting.')
}
```

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@eco-foundation/routes",
-  "version": "2.1.12",
+  "version": "3.2.6",
```
MSevey: 🟡 This should not point at `3.x` yet — the go-live baseline is `v2.8.18` (the `v3.x` tags are the never-released phantoms being pruned). Leaving `3.2.6` here means the first real release rewrites `package.json` down into the `2.8.x` line (or to `3.0.0` on the breaking commit), a numeric "downgrade" from `3.2.6` that reads as a regression in history. Cosmetic (no npm publish), but set it to `2.8.18` in this PR so the pre-release state matches the baseline the release flow will actually compute from.

scripts/release/update-versions.ts
carlosfebres: Added in bbb67de7 as an exported assertCanonicalFileUpdated() (kept separate from the zero-check so it's unit-testable): release prepare now throws if contracts/libs/Semver.sol specifically is absent from the rewrite list, with tests covering both the accept and reject paths (9/9 passing). Confirmed your failure trace — test provers alone matching would compile clean and ship a stale deployed version; the guard now closes exactly that path.

package.json
carlosfebres: Set to 2.8.18 in bbb67de7 — matches the go-live baseline now that the 3.x phantoms are pruned (pruning completed yesterday; v2.8.17 is the highest tag in the repo).

MSevey (approved):

PR #414, opened 2026-07-14
chore(release): 2.10.0 [skip ci]

carlosfebres:
Release commit for v2.10.0, prepared by semantic-release locally (the workflow's direct push to main is blocked by the org ruleset — bypass fix pending with an org owner).

Contains exactly what the release workflow would have committed:
- `version()` rewritten to 2.10.0 in Semver.sol + the two test provers
- `package.json` → 2.10.0
- `CHANGELOG.md` entry for 2.10.0 (feat #409)

After squash-merge, the `v2.10.0` tag + GitHub release will be created on the merge commit manually. `[skip ci]` in the squash title keeps the release workflow from re-firing on this commit.


CHANGELOG.md
MSevey, line 1: I don't think we need to keep this file. if we are doing actually github releases, the changelog can be auto generated there.

MSevey: fine as a follow up, not blocking

MSevey (approved):

PR #415, opened 2026-07-14
ci: release-PR bridge (release without pushing to main)

carlosfebres:
## Summary

Replaces the push-to-main release flow (shipped in #410) with a **release-PR bridge**. The org-level "Default branch protection" ruleset forbids direct pushes to `main` with no bypass actors, and the team does not want to weaken that org-wide setting — so releases now flow through a normal, approved PR, the same mechanism the ruleset already mandates.

This was proven necessary the hard way: after #410 merged, the release workflow's push-back was rejected (GH013), and v2.10.0 had to be cut by hand through a PR.

## How it works

1. **`release-pr.yaml`** (push to `main`): computes the next version (`semantic-release --dry-run`), rewrites contract `version()` + `package.json` via the existing `scripts/release/update-versions.ts`, regenerates `CHANGELOG.md`, and force-pushes an **auto-refreshing** PR from branch `autorelease/next`, titled `chore(release): x.y.z`. Authored with a GitHub App token so required CI checks actually run on it. New commits to `main` while the PR is open refresh its version and notes.
2. A human **approves and squash-merges** the release PR.
3. **`release-tag.yaml`** (push to `main`): detects the merged `chore(release): x.y.z` commit, resolves it by message, verifies `package.json` matches, tags `vx.y.z`, and publishes the GitHub release (PR body as notes, CHANGELOG fallback). A `workflow_dispatch` recovery path re-runs it manually if a squash title is ever mangled.

`autorelease/next` deliberately does **not** match the protected `release*` glob. The baseline guard (stable tag ≥ `v2.8.17` reachable) is preserved.

## What changed

- Added `.github/workflows/release-pr.yaml`, `.github/workflows/release-tag.yaml`; deleted `.github/workflows/release.yaml`.
- Slimmed `.releaserc.json` to analyzer + release-notes-generator + exec (version computation only).
- devDependencies: removed `@semantic-release/{changelog,git,github}`; added `conventional-changelog-cli`.
- Docs: `scripts/README.md` + `CLAUDE.md` Releases sections rewritten.
- **No `contracts/` changes** (`git diff main..HEAD -- contracts/` is empty).

Design: `CLAUDE/specs/2026-07-14-release-pr-bridge-design.md` · Plan: `CLAUDE/plans/2026-07-14-release-pr-bridge.md`

## ⚠️ Prerequisites before the first release (no org ruleset change)

1. **GitHub App** (org owner): create `eco-release-bot` — permissions Contents read/write + Pull requests read/write, installed on `eco/eco-routes` only. Add repo secrets `RELEASE_BOT_APP_ID` and `RELEASE_BOT_PRIVATE_KEY`. This is a repo secret, **not** a ruleset bypass — needed because the default `GITHUB_TOKEN` can't trigger CI on the release PR.
2. **Repo admin**: confirm squash-merge commit message defaults to the PR title (Settings → General → Pull Requests).
3. After merge: close the stopgap release PR #414 if still open — the first `release-pr.yaml` run regenerates the v2.10.0 release PR through the new machinery.

## Notes

- Two CI-only defects were caught in final review (both invisible to local `--no-ci` rehearsals) and fixed: the workflow needed `contents: write` for semantic-release's CI auth probe (the ruleset still blocks real pushes), and the detection regexes now tolerate the `(#NNN)` suffix GitHub adds to squash-merge commit subjects.
- Version rewriting still changes bytecode and CREATE2 addresses on each release (unchanged from the prior design).

## Test plan

- [x] `yarn build`, `yarn test:ts` (9/9), `forge test` (591/591) all green on the branch
- [x] Both workflows parse; version-detection regexes accept plain + `(#NNN)` forms and reject non-release/`[skip ci]` subjects
- [x] Local rehearsal: version computed, rewrite + `forge build` succeed, clean revert
- [x] Recovery path resolves the release commit by message (not branch tip) with a scratch-repo test; `workflow_dispatch` version input validated as semver
- [ ] After GitHub App + secrets land: first live `release-pr.yaml` run opens the release PR with CI green; merge triggers `release-tag.yaml` to tag + publish


CLAUDE/plans/2026-07-14-release-pr-bridge.md
MSevey, line 1: did you mean to include this?

CLAUDE/specs/2026-07-14-release-pr-bridge-design.md
MSevey, line 1: I think having a `RELEASE.md` in the root of the repo is more standard for documenting the release process for a repo.

.github/workflows/release-tag.yaml
```diff
@@ -0,0 +1,114 @@
...
+        with:
+          fetch-depth: 0 # needed to locate the release commit by message
+
+      - name: Resolve release commit
+        if: steps.detect.outputs.version != ''
+        id: resolve
+        env:
+          VERSION: ${{ steps.detect.outputs.version }}
+        run: |
+          SHA=$(git log --fixed-strings --grep="chore(release): $VERSION" --format=%H -n 1)
```
MSevey: **Resolve the release commit by push SHA on the push path, not a message grep.**

`git log --fixed-strings --grep="chore(release): $VERSION" -n 1` searches all history and takes the newest *message* match — and `--grep` matches commit **bodies** too. On the `push` trigger this is harmless (the release commit is the tip), but on the `workflow_dispatch` recovery path a later commit whose body merely contains `chore(release): x.y.z` can become the newest match. It then either fails the anchored subject re-check below (safe-fail, but blocks the recovery you dispatched for) or, worst case, resolves an unintended commit.

The workflow already knows the exact commit on the push path: `${{ github.sha }}`. Suggest tagging that directly when not dispatched, and keeping the grep only as the dispatch fallback.

.github/workflows/release-pr.yaml
```diff
@@ -0,0 +1,136 @@
...
+          # uses for the new section header it prepends to CHANGELOG.md
+          npx conventional-changelog -p angular -i CHANGELOG.md -s
+          npx conventional-changelog -p angular > .release-notes.md
+          echo "--- release notes ---"
+          cat .release-notes.md
+
+      - name: Mint App token
+        if: steps.skip.outputs.skip == 'false' && steps.version.outputs.version != '' && env.DRY_RUN != 'true'
+        id: app-token
+        uses: actions/create-github-app-token@v1
```
MSevey: **Pin the token-minting action to a full commit SHA.**

This is the one step that consumes `RELEASE_BOT_APP_ID` / `RELEASE_BOT_PRIVATE_KEY` (push-to-branch + PR write), yet it floats on the mutable `@v1` tag while every other action here is pinned — `checkout@v4.2.2`, `setup-node@v4.1.0`, and `pr-title-check` pins by SHA. A moved `v1` tag would run unreviewed code in the context that holds the release-bot private key. Pin `actions/create-github-app-token` to a specific commit SHA (it's the highest-value step in the pipeline).

MSevey (commented):
Two robustness/supply-chain notes from a pre-landing review. Neither is a blocker given `main`'s required status checks, but both are worth addressing before the first live release. The injection surface, token-context isolation (mint on `push`/`dispatch` only), no-double-trigger behavior, and the `--dry-run` version-computation pipeline all reviewed clean.

.github/workflows/release-pr.yaml
carlosfebres: Pinned to `actions/create-github-app-token@d72941d797fd3113feb6b93fd0dec494b13a2547 # v1.12.0` in 59b2cfc9 — confirmed that SHA is the v1.12.0 tag commit. Agreed it's the highest-value step to pin.

.github/workflows/release-tag.yaml
carlosfebres: Fixed in 59b2cfc9: on the `push` trigger the workflow now tags `${{ github.sha }}` directly (the release commit is the pushed tip), and the message grep is kept only as the `workflow_dispatch` recovery fallback. The anchored subject re-check still guards the dispatch path.

CLAUDE/specs/2026-07-14-release-pr-bridge-design.md
carlosfebres: Done in d28970cd: added a root RELEASE.md as the canonical release-process doc; scripts/README.md and CLAUDE.md now point at it.

CLAUDE/plans/2026-07-14-release-pr-bridge.md
carlosfebres: Removed from the PR in d28970cd (along with the design spec) — the internal plan/spec docs net out of the diff now; RELEASE.md is the doc that stays. They remain in git history if ever needed.

MSevey (approved):

PR #416, opened 2026-07-15
fix(intent-source): refund excess native in fundFor/publishAndFundFor (Octane V1)

carlosfebres:
## V1 — Refund excess native in `fundFor` / `publishAndFundFor`

**Octane severity:** High · **Reassessed:** Medium

### Problem
The permit/allowance "For" funding path forwards the **entire** `msg.value` into the vault and never refunds the excess, unlike `fund`/`publishAndFund` which transfer only the missing native via `_fundNative` and then call `Refund.excessNative()`. Consequences:
- If the intent is already `Funded`, `onlyFundable` returns early and the sent ETH is **stranded on the Portal**, where the next `fund`/`publishAndFund` caller sweeps the whole Portal balance via `Refund.excessNative()` — an anyone-can-take path.
- Any surplus above the remaining native requirement is **trapped in the vault** and later swept to `reward.creator` on refund, not the overpaying caller.

### Fix
- `_fundIntentFor`: `_fundNative(vault, reward.nativeAmount)` then `IVault(vault).fundFor{value: 0}(...)` — caps the vault transfer to the missing amount.
- `fundFor` and `publishAndFundFor`: call `Refund.excessNative()` as the final statement, mirroring the funded path.

### Validation
`test/source/IntentSource.t.sol` — `testFundForRefundsExcessNative`, `testPublishAndFundForRefundsExcessNative`, `testFundForAlreadyFundedDoesNotStrandNative`. Full IntentSource suite green (59/59).

### Boundaries
The whole-balance sweep in `Refund.excessNative()` is existing, intended design (Portal holds no resting ETH); this PR removes the only identified way ETH became stranded there.

---


contracts/IntentSource.sol
```diff
@@ -734,7 +736,13 @@ abstract contract IntentSource is OriginSettler, IIntentSource {
...
 
         vault = _getOrDeployVault(intentHash);
-        bool fullyFunded = IVault(vault).fundFor{value: msg.value}(
+
+        // Cap the native contribution to the amount still missing from the vault (mirrors the
+        // funded path in {_fundIntent}); the vault balance is what {fundFor} checks against
+        // `reward.nativeAmount`, so no native is forwarded into the vault call itself.
+        _fundNative(vault, reward.nativeAmount);
+
+        bool fullyFunded = IVault(vault).fundFor{value: 0}(
```
fish-sammy:
```suggestion
        bool fullyFunded = IVault(vault).fundFor(
```

fish-sammy: Also, `fundFor` no longer needs to be `payable` after this fix.

carlosfebres: Done — `IVault.fundFor` is now non-payable and returns nothing. Native is pre-funded to the vault via `_fundNative` before the call, so `fundFor` just pulls the token legs, and `_fundIntentFor` derives `fullyFunded` from `_isRewardFunded(reward, vault)` read after the call (a reentrant permit can't fake the post-state balances). Pushed as 3c881c85.

Note on merge order: this makes `IVault.fundFor` non-payable, while #420 makes the same method void but keeps `payable` (it still forwards msg.value). #420 should merge first; this branch then rebases and drops `payable`. If this lands first, #420's `fundFor{value: msg.value}` call won't compile.

carlosfebres: Superseded by the non-payable + void refactor — the call is now `IVault(vault).fundFor(reward, funder, IPermit(permitContract));` with no value and no captured return; `fullyFunded` comes from `_isRewardFunded` afterwards.

fish-sammy (dismissed):

fish-sammy (approved):

PR #417, opened 2026-07-15
feat(erc7683): accept ERC-1271 signatures in gasless openFor (Octane V11)

carlosfebres:
## V11 — Accept ERC-1271 signatures in gasless `openFor`

**Octane severity:** Informational · **Reassessed:** Informational

### Problem
`OriginSettler._validateOrderSig` uses `ECDSA.recover` with strict equality to `order.user`, so ERC-1271 contract wallets (Safe, ERC-4337) cannot use the gasless `openFor` path.

### Fix
Verify via OpenZeppelin `SignatureChecker.isValidSignatureNow(order.user, digest, signature)` — identical acceptance for EOAs, plus ERC-1271 support. Replay protection is unchanged (guarded by vault state in `_publishAndFund`, not the signature scheme).

### Validation
`test/interfaces/OriginSettler.t.sol` — ERC-1271 valid/invalid and EOA valid/invalid cases via the real `Portal.openFor` path. New `contracts/test/MockERC1271Wallet.sol`.

### Boundaries
Forward-looking enhancement, not a security patch; the non-gasless `open()` path already works for contract wallets.

---


MSevey (dismissed):
Approving — the change is sound.

Verified:
- **EOA behavior unchanged.** OZ 5.0.2 `SignatureChecker.isValidSignatureNow` tries `ECDSA.tryRecover` first; for an EOA the ERC-1271 fallback staticcalls a code-less address and returns false, so validity is still `recovered == order.user`. High-s malleable sigs stay rejected.
- **Digest binds all funding-relevant fields** (`order.user`, nonce, chain id, deadlines, and `keccak256(orderData)` which encloses reward/creator/prover/tokens/route).
- **Replay protection stays state-based** — `rewardStatuses[intentHash]` via `onlyFundable`, consulted only at the escrow step and never cached, so ERC-1271 revocability can't strand funds or enable a second fund/withdraw.
- **No new reentrancy** — ERC-1271 validation is a `staticcall`.
- All 4 new tests pass through the real `Portal.openFor` path.

**One assumption to document — @carlosfebres:** this widens the trust boundary. After this change, any contract at `order.user` is trusted to gate its own signatures. A *permissive or buggy* ERC-1271 wallet (one that returns the `0x1626ba7e` magic value for signatures its owner didn't strictly authorize) that has also approved the Portal could be escrowed-from by a third party, with the refund routed to an attacker-chosen `reward.creator` (creator ≠ user is permitted, and is itself inside the signed `orderData`). This is the accepted trade-off of supporting Safe/ERC-4337, not a flaw in `OriginSettler` — but the assumption "the wallet at `order.user` implements ERC-1271 correctly and strictly" should be written down.

I'd like that context to live **with the code**, not just in this PR description, so it survives past the PR. Where do you think it belongs — a `@dev` NatSpec note on `_validateOrderSig` (and/or `openFor`), a trust-assumptions section in `SECURITY.md` / docs, or somewhere else? Your call on the exact spot; just want the assumption captured somewhere durable before this lands.

carlosfebres:
Documented the ERC-1271 trust assumption in-code as you asked, @MSevey — put it as a `@dev` NatSpec block on `_validateOrderSig` (the site that actually widens the trust boundary), with a short pointer note on `openFor`. It captures the four points you verified: (1) any contract at `order.user` is trusted to gate its own signatures; (2) a permissive/buggy ERC-1271 wallet that has approved the Portal can be escrowed-from by a third party with the refund routed to an attacker-chosen `reward.creator` (creator ≠ user is permitted and is inside the signed `orderData`); (3) this is the accepted Safe/ERC-4337 trade-off, not a flaw; (4) EOA behavior is unchanged and high-s malleable sigs stay rejected. Pushed as c573b83a. Documentation only, no behavior change.

contracts/ERC7683/OriginSettler.sol
```diff
@@ -168,10 +180,18 @@ abstract contract OriginSettler is IOriginSettler, EIP712 {
...
-        address signer = hash.recover(signature);
+        bytes32 digest = _hashTypedDataV4(structHash);
 
-        return signer == order.user;
+        // Use SignatureChecker so both EOA (ECDSA) signatures and ERC-1271
+        // contract-wallet signatures (e.g. Safe) are accepted on the gasless
+        // openFor path. For EOAs this is equivalent to the previous
+        // ECDSA.recover + equality check.
+        return
+            SignatureChecker.isValidSignatureNow(
```
MSevey: **Trust boundary: the `openFor` funding path now accepts a contract's own say-so**

Worth making explicit because it's a real widening, not a nit. Today `_validateOrderSig` is `ECDSA.recover(...) == order.user`, which a contract address can never satisfy — so a Portal-wide ERC-20 approval is currently spendable only by the owner via `open()`, or with their ECDSA signature via `openFor()`. (I checked the other funder-argument entry points: `fundFor` / `publishAndFundFor` route through `Vault.fundFor` → `_fundFrom`, which reads a **vault-scoped** allowance, so they can't reach a Portal approval. That separation looks deliberate and correct.)

After this change, a Portal approval becomes spendable on the say-so of whatever contract sits at `order.user`. The NatSpec block already names the attacker-chosen `reward.creator`, which is the right instinct. Two things it doesn't say:

1. `reward.deadline` is also attacker-chosen and can be set in the past, so `_validateRefund` passes immediately — the exposure isn't "escrowed now, refunded later," it's same-transaction.
2. There's no containment offered, only documentation.

**Suggestion:** bind `reward.creator == order.user` on the `openFor` path. It's cheap, it doesn't affect the EOA flow, and it downgrades the failure mode from theft to griefing if a wallet's ERC-1271 policy turns out to be permissive. Doesn't close everything, but it's the highest value-per-line change available here.

Genuinely a judgment call whether it's worth it — depends on whether permissive contract wallets are realistic in your user population. Flagging so it's a decision rather than an omission.

MSevey (commented):
Reviewed at `c573b83a`. The change does what it says: I ran the full suite (595 pass, 0 fail) and probed the rejection paths directly — `address(0)`, a contract with no `isValidSignature`, a wallet that reverts, a wallet returning a non-magic `bytes4`, an empty signature, and high-s malleability all reject correctly. The high-s claim in the new NatSpec at line 161 is accurate; I tested it specifically. EOA behavior is unchanged, and the happy path short-circuits on the `||` in `SignatureChecker`, so there's no gas regression for the common case.

Two inline comments on the trust boundary, plus two smaller things here:

**Prettier.** `contracts/ERC7683/OriginSettler.sol:190` and `test/interfaces/OriginSettler.t.sol:431` fail `prettier --check`. Both files were clean at the merge base, so this PR introduces them and `yarn format` is a CI gate. One `npx prettier --write` on the two files fixes it.

**Test coverage.** The four new tests cover the happy path and one wrong-key case. The six negative cases listed above aren't covered, including the high-s one the NatSpec asserts explicitly. I've written and verified all six passing — happy to hand them over. Two smaller notes: `_buildGaslessOrder` hardcodes `creator == user`, which hides the boundary that matters most, and `_fundAndApprovePortal` uses `vm.prank(wallet)` to approve rather than executing through the wallet.

Separately, I filed PAR-402 for a pre-existing `nonce`/duplicate-event issue found during this review — not introduced by this PR, so it isn't a blocker here.

contracts/ERC7683/OriginSettler.sol
MSevey: **Unbounded staticcall to an attacker-controlled address**

`SignatureChecker.isValidSignatureNow` reaches OZ 5.0.2 `SignatureChecker.sol:41-44`, which does `signer.staticcall(...)` with no gas cap and materializes the full returndata into `bytes memory result`. `order.user` is unvalidated calldata — `openFor` checks deadline, settler, chainId and orderDataType, never `user`.

I measured it: a contract returning 4 MiB from `isValidSignature` burns **68,338,759 gas** in a single `openFor` call, and it's tunable upward without bound since memory expansion is quadratic.

Before this change it was structurally impossible — the old `hash.recover(signature)` made zero external calls. The exposed party is exactly who the gasless path exists for: a relayer paying gas for someone else's order. Attacker cost is one small contract deploy; relayer cost is a full-gas-limit OOG revert, repeatable.

No fund loss, and simulation mitigates it. I'd **not** recommend hand-rolling bounded-staticcall assembly into this contract — the cure carries more risk than the disease. But a line in the NatSpec so relayer operators know simulation is load-bearing seems worth it.

Also worth adding to the same NatSpec block while you're in there: ERC-1271 signatures are **revocable** (OZ's own note at `SignatureChecker.sol:23-24`). A Safe rotating owners flips a valid order to invalid between simulation and inclusion — an EOA signature can never do that, and it's a new liveness consideration for solvers.

carlosfebres:
Thanks for the thorough pass, @MSevey — all addressed in c77b44b9 (after merging main):

- **Prettier**: fixed; `prettier --check` is clean on both files now.
- **Creator-binding**: merging main pulled in #421, so `openFor` now enforces `reward.creator == order.user`. That's exactly the containment you suggested — it downgrades a permissive-wallet failure from **theft to griefing** (funder and refund recipient are the same wallet). I rewrote the trust-assumption NatSpec around that, and added your other two points: the attacker-chosen `reward.deadline` (same-transaction refund, still back to the wallet) and the unbounded-staticcall DoS on relayers (I kept your call — no hand-rolled bounded staticcall; documented that simulation is load-bearing), plus ERC-1271 revocability as a solver liveness note.
- **Negative tests**: went ahead and wrote the six myself so the PR is self-contained — `address(0)` user, a code-bearing non-wallet, a reverting ERC-1271 wallet, a wrong-magic wallet, an empty signature, and a high-s malleable signature. All assert `InvalidSignature`. Suite is 20/20. If your versions cover angles mine miss, happy to swap or add.
- On the helper notes: with #421's binding, `_buildGaslessOrder`'s `creator == user` is now the only valid shape (mismatch reverts `InvalidCreatorBinding`, covered by your #421 test), and `_fundAndApprovePortal` uses `vm.prank` because `MockERC1271Wallet` has no execution entrypoint — open to a richer mock if you'd prefer.
- Filed-separately **PAR-402** noted, not touched here.

Note the branch now merges main, so it's no longer behind/dirty.

fish-sammy (approved):

MSevey (approved):

PR #418, opened 2026-07-15
fix(prover): failure-tolerant native refunds; refund trapped Polymer value (Octane V9)

carlosfebres:
## V9 — Deliver native refunds to smart-account recipients; revert on failure; refund trapped Polymer value

**Octane severity:** Low · **Reassessed:** Low

### Problem
- `MessageBridgeProver._sendRefund` used `transfer()` (2300-gas stipend). `Inbox.prove` forwards the Portal's **entire** balance to the prover, so forced dust or fee-quote overpayment yields a nonzero refund that reverts for smart-account / EIP-7702 solvers whose `receive` needs >2300 gas — DoS-ing `prove`/`fulfillAndProve`.
- `PolymerProver.prove` is `payable` but never consumed/refunded forwarded ETH and has no withdrawal path, so it was permanently trapped.

### Fix
Both refund sites use an uncapped low-level `call{value: amount}("")` — forwarding all remaining gas rather than the 2300-gas `transfer`/`send` stipend, so a smart-account / EIP-7702 recipient can receive the refund. On a failed transfer the call **reverts** with `RefundFailed(recipient, amount)` instead of swallowing the failure. The refund recipient is always the tx caller (`Inbox.prove` passes `msg.sender`), so reverting surfaces a genuinely-unpayable caller loudly rather than silently stranding their ETH as dust with no sweep path — and it can only self-DoS that caller, with no third-party griefing vector.

- `MessageBridgeProver._sendRefund`: uncapped `call` + `revert RefundFailed(recipient, amount)` on failure. `RefundFailed` is declared in `IMessageBridgeProver` (shared by Hyper/Meta/CCIP/LayerZero).
- `PolymerProver.prove`: refund forwarded `msg.value` to the caller with the same uncapped `call` + revert (local `RefundFailed`).

Reentrancy is contained: the refund is the terminal statement of `prove`, `Inbox.prove` is `nonReentrant` and holds the guard for the whole call, and it drains the Portal's balance into the prover before the refund callback — so a reentrant `prove` is guard-blocked and carries 0 value.

### Validation
- `test/prover/HyperProver.t.sol` — `testProveRevertsWhenRefundRecipientRejects` (rejecting recipient → `RefundFailed`); `testProveRefundDeliveredToGasHungryRecipient` (smart-contract recipient needing >2300 gas still receives).
- `test/prover/PolymerProver.t.sol` — `testProveRevertsWhenRefundRecipientRejects`; `testProveRefundsForwardedValueToSender`; `testInboxProveReentrancyIsBlocked`.
- `test/prover/LayerZeroProver.t.sol` — parity: rejecting-recipient revert + gas-hungry delivery.
- 162 prover / 598 full `forge test` pass; `prettier --check` clean.

---
_Source: Octane automated analysis (2026-07-11), independently validated against the code at `b3eb0b35`. Severity reassessed during validation. Refund approach revised 2026-07-30 per internal review: revert on an undeliverable refund while keeping uncapped gas so smart-account recipients can still receive._


MSevey (dismissed):
LGTM — approving. The fix is correct, minimal, consistent with the repo's existing idiom (`LocalProver.sol:237` and the `Refund` lib already use the same failure-tolerant `call{value:}` pattern), and well-tested (both new regression tests + all 157 prover tests pass).

Two things I verified while reviewing that aren't defects, but that I'd like to see captured in-code so future audits/reviews don't re-flag them. **@carlosfebres — do you think it'd be valid to fold this context into the existing comments where applicable?** Documenting that we understand the flow here saves a round-trip on the next audit pass.

**1. The 2300-gas reentrancy barrier is intentionally gone.** Switching `transfer()` → `call{value:}` forwards all gas, so the refund recipient can now reenter. I checked and it's not exploitable: the refund is the *terminal* statement in both `MessageBridgeProver.prove` and `PolymerProver.prove` (no post-refund state), `Inbox.prove` has already drained `address(this).balance` before the callback so a reentrant `prove` forwards 0 value, and source-side settlement is idempotent via reward status. Worst case is a caller re-emitting `IntentProven`/`IntentFulfilledFromSource` at their own gas expense. Worth one line noting "refund is terminal + balance already drained, so reentrancy is non-exploitable" next to each refund.

**2. A rejected refund leaves dust stuck with no sweep path.** If the recipient reverts, the ETH stays in the prover permanently — there's no `rescue`/`withdraw` on `MessageBridgeProver` or `PolymerProver` (the `HyperProver` regression test asserts this retention). That's the correct trade (DoS avoidance > dust recovery), and since the recipient is the caller it's self-inflicted. A short note that failed refunds are intentionally retained-as-dust rather than recoverable would make the trade explicit.

Neither is blocking — approving as-is.

carlosfebres:
Folded both points into the existing refund comments as you suggested, @MSevey. `MessageBridgeProver._sendRefund` and `PolymerProver.prove` now each note (1) the refund is the terminal statement and the Portal balance is already drained by `Inbox.prove`, so the all-gas `call` can't be exploited via reentrancy, and (2) a rejected refund is intentionally retained as dust — no rescue path, and the loss is self-inflicted since the recipient is the caller. Pushed as 5a15eca2. Comments only, no bytecode change. (Corrected one phrasing so both provers consistently say the *Portal's* balance is drained, not the prover's.)

contracts/prover/MessageBridgeProver.sol
```diff
@@ -64,7 +64,20 @@ abstract contract MessageBridgeProver is
...
+        // Dropping transfer()'s 2300-gas cap forwards all gas, so the recipient
+        // can reenter — but this call is the terminal statement of prove() (no
+        // post-refund state) and Inbox.prove has already drained the Portal's
+        // balance before the callback, so a reentrant prove() carries 0 value. Nothing
+        // is left to exploit. If the recipient rejects the transfer the ETH is
+        // intentionally retained as dust (there is no rescue/sweep path): DoS
+        // avoidance is worth more than dust recovery, and the recipient is the
+        // caller, so the loss is self-inflicted.
+        (bool _ok, ) = payable(recipient).call{value: amount}("");
+        _ok;
```
MSevey: **[hardening, non-blocking]** The reentrancy safety argument is sound but it's comment-enforced, and it's documented in the wrong file.

I verified the invariant holds today across every path (details in the review summary). The issue is durability: it's *asserted* here in `contracts/prover/`, but it's *enforced* by code in `Inbox.sol` and `ERC7683/DestinationSettler.sol`. There's no reentrancy guard anywhere on Portal or Inbox — `ReentrancyGuard` appears only in `deposit/BaseDepositAddress.sol` and `prover/LocalProver.sol`.

So a future PR that adds a statement after `prove(...)` in `Inbox.prove` or `fulfillAndProve` silently breaks a live security property, and nothing fails: no guard, no test, and this comment still reads as if it were true.

Dropping the 2300-gas cap is the right call — but it converts a structural protection into a documentary one, and the doc lives two directories away from the code that upholds it.

Suggest one of:
1. Add `nonReentrant` to `Inbox.prove` (OZ `ReentrancyGuard` is already a dependency). Makes it structural again.
2. At minimum, mirror the invariant as a comment at `Inbox.sol:195`, where the enforcing code actually lives, so the next person editing that function sees it.

test/prover/HyperProver.t.sol
```diff
@@ -523,6 +523,50 @@ contract HyperProverTest is BaseTest {
...
 
+    /**
+     * @notice prove() must not revert when the refund recipient rejects ETH
+     * @dev Regression test for V9: _sendRefund previously used transfer() with a
+     *      2300-gas stipend, so an overpayment refund to a smart-account /
+     *      EIP-7702 wallet whose receive() reverts (or consumes >2300 gas) would
+     *      DoS prove()/fulfillAndProve(). The failure-tolerant low-level call
+     *      makes the refund non-reverting; the excess simply stays in the prover.
+     */
+    function testProveRefundDoesNotRevertOnRejectingRecipient() public {
```
MSevey: **[test coverage]** This covers a different failure mode than the one the PR describes.

The stated motivation is "smart-account / EIP-7702 solvers whose `receive` needs >2300 gas." `RejectingRefundRecipient` **reverts**. Those aren't the same case:

- A receiver that **reverts**: fails under `transfer()`, and under `call` it still fails — we just tolerate it. Behavior changes from *revert* to *tolerated failure*.
- A receiver that **accepts but burns >2300 gas**: OOG-reverts under `transfer()`, and **succeeds** under `call`. Behavior changes from *failure* to *success*.

The second is the actual production scenario, and it's the stronger assertion — it proves the refund is *delivered*, not merely that `prove()` didn't revert. Right now nothing tests that a legitimate smart-account solver actually receives their money.

~15 lines to add:

```solidity
contract GasHungryRefundRecipient {
    uint256 private slot0;
    uint256 private slot1;
    receive() external payable {
        // >2300 gas: two cold SSTOREs
        slot0 = block.number;
        slot1 = block.timestamp;
    }
}
```

Then assert the recipient's balance actually increases by the refund.

test/prover/HyperProver.t.sol
```diff
@@ -523,6 +523,50 @@ contract HyperProverTest is BaseTest {
...
+            encodeProofs(intentHashes, claimants),
+            proverData
+        );
+
+        // Refund could not be delivered, so it is retained by the prover
+        // (fee = 100000 wei was consumed by the mock mailbox dispatch).
+        assertEq(address(recipient).balance, 0);
+        assertEq(
+            address(hyperProver).balance,
+            proverBalanceBefore + overpayment - 100000
```
MSevey: **[nit]** Magic number — `TestMailbox.FEE` is already a `public constant` (`contracts/test/TestMailbox.sol:30`).

Use `mailbox.FEE()` instead of hardcoding `100000`. If the mock fee ever changes, this fails as an opaque arithmetic mismatch pointing at the refund logic rather than at the mock.

test/prover/PolymerProver.t.sol
```diff
@@ -212,6 +212,39 @@ contract PolymerProverTest is BaseTest {
...
 
+    /**
+     * @notice prove() must refund forwarded ETH to the caller rather than trap it
+     * @dev Regression test for V9: Polymer proving requires no bridge fee, but
+     *      Inbox.prove forwards the Portal's entire balance (forced dust or
+     *      overpayment). Previously prove() was payable and only emitted an event,
+     *      permanently trapping any msg.value in the prover. It must now refund
+     *      the forwarded value to the sender.
+     */
+    function testProveRefundsForwardedValueToSender() public {
```
MSevey, line 224: **[test coverage]** Two gaps on the Polymer side.

1. **No rejecting-recipient test.** `PolymerProver.prove` has the same failure-tolerant refund branch as `_sendRefund`, but this test only uses `creator`, an EOA that always accepts. HyperProver got `RejectingRefundRecipient`; Polymer didn't. The dust-retention branch here is entirely unexercised — worth mirroring, since it's the branch that silently keeps funds.

2. **No integration coverage.** This calls `polymerProver.prove{value: ...}` directly with `vm.prank(address(portal))`. The behavior the PR is actually fixing is `Inbox.prove` forwarding `address(this).balance` into the prover — and that path isn't covered in either test file. A test that funds the Portal and goes through `Inbox.prove` would catch regressions in the forwarding itself, which is where the value originates.

MSevey (commented):
## Review: correct change, incomplete coverage of its own bug class

The fix itself is right and the tests are real. I verified rather than assumed:

- `forge build` compiles with **no new warnings** — the `(bool _ok, ) = ...; _ok;` idiom silences both 9302 and 2072. The two remaining unchecked-call warnings are pre-existing (`Vault.sol:127`, `LocalProver.sol:237`).
- All 157 prover tests pass.
- I copied both new tests onto the base commit `b3eb0b35` and confirmed they **genuinely fail there** (`RejectingRefundRecipient: I reject your ETH`, and `10e18 != 11e18`). These are real regression tests, not tautologies.
- The "terminal statement / no post-refund state" claim in the new comments **holds today** on every path. I checked all of them: `Inbox.prove:196` is the last statement; `fulfillAndProve:131` is followed only by `return result`; `DestinationSettler.fill:41` has nothing after it; `MessageBridgeProver.prove:146` ends with `_sendRefund`; and none of HyperProver / MetaProver / LayerZeroProver / CCIPProver override `prove`. At callback time `claimants[intentHash]` is already written (`Inbox.sol:242`, before the external call) and the executor has already run. No half-updated state, not exploitable.

Four inline comments below — one hardening suggestion and three on test coverage. None of them block.

**One thing worth raising before this lands as the complete V9 fix:** the same trapped-value shape this PR fixes for Polymer exists on another prover, in a form with a worse outcome than Polymer's. I'm routing that one through private disclosure rather than describing it here, per the repo's security policy — @MSevey has the details.

Also: `Inbox.prove` forwarding `address(this).balance` (rather than a scoped amount) is the root of this whole class. It's out of scope here, but it deserves its own issue rather than another per-prover patch.

_Reviewed in an isolated worktree at `5a15eca2`._

carlosfebres:
Addressed the review in 17ce2d11, @MSevey:

**Hardening (your option 1):** added `ReentrancyGuard` to `Inbox` and marked `Inbox.prove` `nonReentrant`, so the reentrancy safety is now structural rather than only documented in the prover. Verified `fulfillAndProve`→`prove` doesn't deadlock (it calls `prove` without holding the guard, so the internal call is the first entry) — full prover + prove/fulfill suites green (160 + 78).
- ⚠️ **Storage layout**: inheriting `ReentrancyGuard` adds a slot, so Portal's storage layout changes. Fine for the next CREATE2 deploy (new address), but flagging that it's not layout-compatible with the current deployment.

**Tests:**
- `testInboxProveReentrancyIsBlocked` — goes through `Inbox.prove` (via Polymer, no fee → full refund to the caller), reenters from the recipient's `receive()`, and asserts the reentrant call reverts with `ReentrancyGuardReentrantCall` while the outer `prove()` still succeeds. It fails without the guard (Portal is already drained, so the reentrant call would otherwise carry 0 value and succeed).
- `testProveRefundDeliveredToGasHungryRecipient` — your point exactly: a recipient that accepts but burns >2300 gas (two cold SSTOREs); asserts the refund is *delivered* (`overpayment - mailbox.FEE()`), not just tolerated.
- `testProveRefundRetainedWhenRecipientRejects` (Polymer) — exercises the dust-retention branch that was unexercised on the Polymer side.
- `mailbox.FEE()` replaces the hardcoded `100000`.

On the **root cause** (`Inbox.prove` forwarding `address(this).balance` rather than a scoped amount): agreed it deserves its own issue rather than more per-prover patches — happy to file it. And understood on the **other prover's trapped-value issue** — leaving that to your private-disclosure routing, not touching it here.

fish-sammy (changes requested):
I don't get this PR. Why do we favor failure-tolerant native refunds over `.transfer`? `Inbox.prove` is permissionless. Anyone can call it and it doesn't have to be the solver. If the solver's contract wallet's receive won't work, it's not the end of the day. Failure-tolerant native refunds mean ETH might get stuck in the prover silently.

carlosfebres:
Heads-up on a design reversal in `4d49b7be` (after an internal review of the refund semantics), since the earlier discussion here was about the failure-tolerant approach.

**Change:** the prover refunds no longer swallow a failed transfer — they now **revert** with `RefundFailed(recipient, amount)`. The gas handling is unchanged: both sites still use an uncapped `call{value}("")` (all gas, not the 2300-gas stipend), so smart-account / EIP-7702 recipients can still receive.

**Rationale:** the refund recipient is always the tx caller (`Inbox.prove` passes `msg.sender`), so reverting surfaces a genuinely-unpayable caller loudly rather than silently stranding their ETH as dust — and it can only self-DoS that caller, no third-party griefing.

**Reentrancy re-checked:** the uncapped call already existed, so the surface is unchanged; the `nonReentrant` guard is held for the whole `prove`, the Portal's balance is drained before the refund callback, and `prove` writes no escrow state — a reentrant `prove` is guard-blocked and value-less. Revert-on-failure is if anything strictly safer (it rolls back any reentrant writes).

**Tests:** the two rejecting-recipient tests now assert the `RefundFailed` revert; `testProveRefundDeliveredToGasHungryRecipient` is kept and still passes (a >2300-gas smart-contract `receive()` is delivered), and LayerZero parity tests were added. 162 prover / 598 full `forge test` pass. Description updated to match.

Note: the branch is 8 commits behind `main`; it'll need a sync before merge.

fish-sammy (approved):

contracts/prover/MessageBridgeProver.sol
carlosfebres: Added `ReentrancyGuard` inheritance + `nonReentrant` on `Inbox.prove` (your option 1) in 17ce2d11 — the guard is enforced now, not just comment-argued.

test/prover/HyperProver.t.sol
carlosfebres: The refund design was since reversed (4d49b7be) to revert on failure instead of swallowing. This test now asserts the `RefundFailed` revert, and a separate gas-hungry-delivery test proves a >2300-gas smart-contract `receive()` still gets paid.

test/prover/HyperProver.t.sol
carlosfebres: Replaced the magic number with `mailbox.FEE()`. 17ce2d11.

test/prover/PolymerProver.t.sol
carlosfebres: Both gaps closed: a Polymer rejecting-recipient test (now asserts the `RefundFailed` revert after the 4d49b7be reversal) and `testInboxProveReentrancyIsBlocked` exercising the `Inbox.prove` forwarding path.

PR #419, opened 2026-07-15
fix(intent-source): treat codeless prover as no-proof so refunds don't brick (Octane W8)

carlosfebres:
## W8 — Treat codeless prover as no-proof so refunds don't brick

**Octane severity:** Medium · **Reassessed:** Low–Medium

### Problem
If a creator sets `reward.prover` to a codeless/reverting address (typo, `address(0)`, or a prover deployed only on another chain), `IProver(reward.prover).provenIntents(intentHash)` reverts. `_validateRefund` calls it **unconditionally, even after the deadline**, so `refund` reverts forever and `recoverToken` cannot rescue reward tokens/native — the escrow is permanently locked.

### Fix
In `_validateRefund`, skip the `provenIntents` call when `reward.prover.code.length == 0` and behave as "no proof" (zero-claimant `ProofData`), so the existing deadline/status branches allow the post-deadline refund. Provers **with** code take the unchanged path.

### Why not the report's suggestion
A blanket `try/catch` would convert a legitimate prover's **transient** revert into "no proof" and let a creator wrongfully refund an already-fulfilled intent (loss shifted to the solver). The deterministic code-length check avoids that. No prover allowlist (would break the permissionless creator-chooses-prover model). The malicious-prover "drain" is out of scope by design (`reward.prover` is committed into `intentHash`).

### Validation
`test/source/IntentSource.t.sol` — `testRefundSucceedsWithCodelessProverAfterDeadline`, `testRefundSucceedsWithZeroAddressProverAfterDeadline`, `testRefundSucceedsWithNormalProverNoProofAfterDeadline` (no regression).

### Boundaries
`withdraw` is intentionally left reverting on a codeless prover (no proof can exist, so no funds are at risk).

---


MSevey (approved):
Approve. Reviewed the diff in isolation against `main`.

**Correctness:** The `code.length > 0` guard in `_validateRefund` correctly treats a codeless prover (`address(0)`, typo, or wrong-chain prover) as "no proof," letting the existing deadline/status branches allow a post-deadline refund instead of reverting forever. Both `refund` and `refundTo` route through `_refund` → `_validateRefund`, so both entry points are covered.

**Tests:** All 20 `*Refund*` tests pass, including the 3 new cases and the solver-protection regressions (`testCantRefundIfCorrectProofExists`, `testRefundRejectsBeforeDeadlineWithoutProof`, `testRefundRejectsWhenIntentNotClaimed`).

**Security:** No exploitable path found (also confirmed by an independent adversarial pass).
- No creator double-spend: a recorded proof requires the prover to have code, and EIP-6780 removes the only way (selfdestruct) to strip code from a prover that already recorded one, so `code.length==0` is a faithful "cannot hold a proof" proxy.
- Solver-protection path is byte-identical for all code-bearing provers.
- `_validateRefund` is `view` and `provenIntents` is `external view` (STATICCALL) — no reentrancy/view concern; the change only removes an external call in the codeless case.
- Leaving `withdraw` reverting on a codeless prover is correct (no claimant can exist, no funds at risk).

The deterministic code-length check is the right choice over `try/catch`, which would risk converting a legitimate prover's transient revert into a wrongful refund.

Minor (pre-existing, not introduced here): the no-proof branch is status-agnostic, but reaching `Withdrawn` first drains the vault, so no double payout is possible.

PR #420, opened 2026-07-15
fix(vault): keep Funded transition status-monotonic vs permit reentrancy (Octane V8)

carlosfebres:
## V8 — Keep `Funded` transition status-monotonic against permit reentrancy

**Octane severity:** Medium · **Reassessed:** Medium

### Problem
`Vault.fundFor` snapshots native completeness **before** the token loop, then calls `permit.transferFrom` on a **caller-supplied, untrusted** permit contract with no reentrancy guard. That callback can reenter public `withdraw`/`refund` while status is still `Initial`, reach a terminal status, and drain the vault; on return, `_fundIntentFor` writes `Status.Funded` **unconditionally**, reverting the terminal state and re-opening a settled intent (a later re-fund could then be withdrawn again).

### Fix
- `Vault.fundFor`: evaluate native completeness **after** the token loop.
- `IntentSource._fundIntentFor`: recompute funding from actual balances via `_isRewardFunded` and only set `Status.Funded` when status is still `Initial`. Same monotonicity guard added to `_fundIntent`.

### Validation
`test/source/IntentSource.t.sol` — `testReentrantRefundDuringFundForKeepsTerminalStatus` (end-to-end reentrancy regression via a malicious permit), `testVaultFundForReturnsFalseWhenNativeShortThoughTokensFunded`.

### Boundaries
Not an unconditional vault drain (a lone attacker round-trips their own tokens); the fix targets the state-machine invariant. A blanket `nonReentrant` guard would be stronger defense-in-depth and can be layered later.

---


contracts/vault/Vault.sol
```diff
@@ -55,7 +55,7 @@ contract Vault is IVault {
         address funder,
         IPermit permit
     ) external payable onlyPortal returns (bool fullyFunded) {
-        fullyFunded = address(this).balance >= reward.nativeAmount;
+        fullyFunded = true;
```
fish-sammy: Portal no longer needs this return value. I'd prefer simplify the implementation of this function and making it not return anything. It's not trustworthy due to the potential re-entrancy draining tokens.

carlosfebres: Done — `IVault.fundFor` now returns nothing and the Vault impl is simplified to just pull the token legs (no in-call completeness flag). `IntentSource` decides funded/partial by reading the vault's balances after the call via `_isRewardFunded`, which a reentrant permit can't fake. Pushed as 251c323f.

While here I also closed a gap the review surfaced: the existing "keeps terminal status" tests all used `nativeAmount = 1 ether`, so the reentrant drain made `fullyFunded` false and the `Status.Funded` assignment was never reached — they'd stay green even without the monotonic guard. Added token-only (`nativeAmount = 0`) reentrant withdraw + refund tests that reach `fullyFunded == true` with a terminal status, so the `== Status.Initial` guard is the only thing preventing an overwrite. Mutation-checked: stripping the guard makes both fail; restoring makes them pass.

carlosfebres:
Heads-up on merge order with #416: both PRs change `IVault.fundFor`. This one (#420) makes it **void** but keeps `payable` (it still forwards `msg.value` to the vault). #416 makes it **void _and_ non-payable** (native is pre-funded there). Please merge **#420 before #416** — #416 then rebases and removes `payable`. If #416 merges first, this branch's `fundFor{value: msg.value}` call won't compile until rebased.

fish-sammy (dismissed):

fish-sammy (dismissed):

test/source/IntentSource.t.sol
```diff
@@ -1573,6 +1575,101 @@ contract IntentSourceTest is BaseTest {
             intent.reward
```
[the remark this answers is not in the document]
carlosfebres: Covered by the four `testReentrant{Refund,Withdraw}{DuringFundFor,TokenOnly}KeepsTerminalStatus` tests. They were also just un-vacuoused (f3283942): a merged Permit2-expiration change (#422) had silently stopped the malicious mocks from reaching the reentrant path, so the mocks now report a live expiration and the reentrancy genuinely fires. Mutation-verified — stripping the `== Status.Initial` guard makes the token-only cases fail (`1 != 3` / `1 != 2`).

fish-sammy (approved):

PR #421, opened 2026-07-15
fix(erc7683): bind gasless order.user to reward.creator in openFor (Octane W2)

carlosfebres:
## W2 — Bind gasless `order.user` to `reward.creator` in `openFor`

**Octane severity:** Medium · **Reassessed:** Low (defense-in-depth)

### Problem
Gasless `openFor` funds from `order.user` but never checks `reward.creator == order.user`, while refunds/recovery always pay `reward.creator`. This is **not** an auth bypass (the EIP-712 signature commits to the full `orderData`, incl. creator/prover) — it reduces to a blind-signature phishing footgun — but aligning the funder with the refund authority is the invariant every existing flow already follows.

### Fix
`openFor`: require `order.user == orderData.reward.creator`, else revert `InvalidCreatorBinding(user, creator)` (new error in `IOriginSettler`). `open()` is untouched (msg.sender funds, creator freely chosen by design).

### Validation
`test/interfaces/OriginSettler.t.sol` — `testOpenForRevertsWhenUserNotCreator` (drives the real `Portal.openFor` signature path). Existing happy paths use `user == creator` and remain valid.

### Boundaries
Does not (and cannot) close the phished-malicious-prover/route legs — prover choice is an explicit protocol trust assumption.

---


MSevey (approved):
Reviewed against `origin/main` (merge-base `b3eb0b35`) in an isolated worktree.

The change is minimal, correct, and well-tested:
- Verified the invariant: `_publishAndFund(..., funder=order.user)` pulls ERC20 rewards from `order.user` while refunds/recovery pay `reward.creator`. The new guard aligns funder with refund authority.
- Guard is placed after signature/type validation and before `_publishAndFund` — signature proven before revert, no state touched.
- `open()` correctly untouched (msg.sender funds, creator freely chosen by design).
- No existing tests break: Hardhat `creates via openFor` uses `user == creator`; the Solidity mock path is unaffected.
- Ran `forge test --match-contract OriginSettlerTest`: 10 passed, 0 failed, including the new `testOpenForRevertsWhenUserNotCreator`.

Non-blocking notes: new public ABI error `InvalidCreatorBinding(address,address)` may want decoding support downstream (routes-ts); gasless orders with `user != creator` will now revert early (previously unsafe since refunds always paid creator).

PR #422, opened 2026-07-15
fix(funding): attempt every reward leg; honor Permit2 expiration (Octane V10)

carlosfebres:
## V10 — Attempt every reward leg; honor Permit2 expiration

**Octane severity:** Low · **Reassessed:** Low

### Problem
- `IntentSource._fundIntent` aggregates `fullyFunded = fullyFunded && _fundToken(...)`; `&&` short-circuits, so after one under-funded leg the later `_fundToken` calls never run and fundable legs are silently skipped under `allowPartial=true` — inconsistent with `Vault.fundFor`.
- `Vault._fundFromPermit` reads the Permit2 allowance amount but discards the expiration, so an expired-but-nonzero allowance reverts `transferFrom` (`AllowanceExpired`) before the standard-ERC20 fallback runs.

### Fix
- `_fundIntent`: evaluate each leg into a local (`bool legFunded = _fundToken(...)`) before aggregating.
- `_fundFromPermit`: read `expiration` and gate the permit transfer on `block.timestamp <= expiration`, falling through to `_fundFrom`.

### Validation
`test/source/IntentSource.t.sol` — `testFundsLaterTokenLegWhenEarlierLegPartial`. `test/core/Vault.t.sol` — `test_fundFor_success_expiredPermitFallsBackToRegularApproval` (+ `MockPermit` harness updated to model real Permit2 expiration semantics).

---


MSevey (approved):
Reviewed in an isolated worktree. Both fixes are correct and minimal.

- `Vault._fundFromPermit`: the `block.timestamp <= expiration` guard is the exact complement of Permit2's own `AllowanceExpired` check, so it skips the permit transfer precisely when Permit2 would revert and falls through to the standard-ERC20 path in `_fundFrom`.
- `IntentSource._fundIntent`: evaluating each leg into a local before aggregating removes the `&&` short-circuit that silently skipped later reward legs after an earlier shortfall; now consistent with `Vault.fundFor`.

Verified: both new tests pass with the fix and genuinely fail against base (`AllowanceExpired(999)`, `0 != 2000`) — real regression tests. Full suites green: Vault 36/36, IntentSource 57/57. No new fund-loss or reentrancy surface introduced.

PR #426, opened 2026-07-27
docs(erc7683): document at-least-once Open/IntentPublished semantics

carlosfebres:
## Problem

`Open` and `IntentPublished` are documented and named as if they were exactly-once per order. They are not.

`orderId` is the intent hash — a pure function of `(destination, routeHash, rewardHash)`. Once an intent reaches `Status.Funded`, `onlyFundable` **returns** rather than reverting, and `_validatePublish` rejects only `Withdrawn`/`Refunded`. So `publish()` re-runs and re-emits on an already-funded intent.

There are three independent doors to a duplicate announcement, not one:

1. **`publish()`** — public, no signature, no deadline. Anyone, forever.
2. **`open()`** — non-gasless, arbitrary caller, no signature, no nonce, no deadline. The caller needs no token balance or allowance, because `onlyFundable` short-circuits before any transfer.
3. **`openFor()`** — `GaslessCrossChainOrder.nonce` is covered by the EIP-712 digest but never consumed on-chain and excluded from `orderId`, so a signature stays replayable until `openDeadline`.

The NatSpec actively misled on this. `OriginSettler` claimed "replay protection" and "protection against replay attacks," and a comment read "No need for replay protection here." All true of *funds*, all false of *events*.

A related consequence: two signed gasless orders differing only by `nonce` resolve to the same `orderId` and the same single escrow.

## Impact

No fund loss and no double-spend. Escrow is single-shot (`onlyFundable` prevents a second pull) and duplicate fills are rejected on the destination chain with `IntentAlreadyFulfilled`.

The exposure is entirely off-chain: any indexer, solver, or quoting service treating `Open`/`IntentPublished` as unique-per-order can double-count, double-quote, or attempt a fill it cannot win. Anyone can trigger it for the cost of gas — and via `publish`/`open`, unbounded in time rather than capped by `openDeadline`.

The misleading NatSpec is the active harm: an integrator trusting "replay protection" builds exactly the non-idempotent consumer that gets hurt.

## What Changed

**Documentation and tests only. No executable line in `contracts/` changed** — every changed line there is a comment. `Portal` deployed bytecode is byte-identical before and after (`sha256 a4544c3d…371b`; the repo builds with `bytecode_hash = "none"` and `cbor_metadata = false`).

- **`contracts/interfaces/ERC7683/IOriginSettler.sol`** — at-least-once contract on `event Open`: both causes, the nonce/`orderId` collapse, and `route.salt` as the way to make genuinely distinct intents.
- **`contracts/interfaces/IIntentSource.sol`** — same on `event IntentPublished`.
- **`contracts/ERC7683/OriginSettler.sol`** — removed the "replay protection" overclaims; rewrote "No need for replay protection here" to state that funds are protected but `Open` is re-emitted, and that this is not specific to the signature path.
- **`contracts/IntentSource.sol`** — documented that `onlyFundable` *returns* on `Funded` (a successful call does not imply funds moved), that `_validatePublish` permits re-publish, and that `publish` is permissionless and repeatable.
- **`README.md`** — new "Event Semantics for Integrators" section.
- **`test/interfaces/OriginSettler.t.sol`** — 7 characterization tests.

## Human Logic

The invariant being documented, not changed:

```text
GIVEN an intent whose rewards are escrowed (Status.Funded)
WHEN  any of open / openFor / publish / publishAndFund is called for that same intent
THEN  the announcement event is emitted again
AND   the escrow is left untouched (no second token pull)
AND   the call succeeds rather than reverting

THEREFORE for consumers:
  orderId / intentHash identifies an order uniquely
  BUT the event carrying it is delivered AT LEAST ONCE, not exactly once
  SO consumers MUST dedupe on orderId and MUST NOT count events as orders

AND separately:
  two gasless orders differing ONLY by nonce
    => identical routeHash and rewardHash
    => identical orderId
    => one intent, one escrow
  to create distinct intents, vary a field feeding the hash (normally route.salt)
```

## Logic Trace

```mermaid
flowchart TD
    A[Intent already Status.Funded] --> B{Entry point}

    B -->|"publish()"| C["no signature<br/>no deadline<br/>permissionless"]
    B -->|"open()"| D["no signature<br/>no nonce, no deadline<br/>no balance or allowance needed"]
    B -->|"openFor()"| E["valid signature required<br/>replayable until openDeadline<br/>nonce signed but never consumed"]

    C --> F["_validatePublish: allows<br/>rejects only Withdrawn / Refunded"]
    D --> F
    E --> F

    F --> G["onlyFundable: RETURNS on Funded<br/>(does not revert)"]
    G --> H[Announcement re-emitted]
    G --> I[Escrow untouched — no second pull]

    H --> J{Consumer dedupes on orderId?}
    J -->|yes| K[Correct — no impact]
    J -->|no| L[Double-count / double-quote / wasted fill attempt]
```

## Service Topology

```mermaid
flowchart LR
    subgraph OnChain["On-chain — source"]
        OS["OriginSettler<br/>open / openFor"]
        IS["IntentSource<br/>publish / publishAndFund"]
        V[("Vault — CREATE2<br/>single-shot escrow")]
    end

    OS -->|"emit Open"| Bus(("Event log"))
    IS -->|"emit IntentPublished"| Bus
    OS --> IS
    IS -->|"onlyFundable — first call only"| V

    subgraph OffChain["Off-chain consumers — where the risk lives"]
        IDX[Indexers]
        SOL[Solvers]
        QT[Quoting services]
    end

    Bus -->|at-least-once| IDX
    Bus -->|at-least-once| SOL
    Bus -->|at-least-once| QT

    IDX -.->|"MUST dedupe on orderId"| Guard{{"documented in this PR"}}
    SOL -.-> Guard
    QT -.-> Guard
```

## Why This Shape

Three options were on the table: (1) document the at-least-once contract, (2) consume the nonce via a per-user mapping or bitmap, (3) make `openFor` revert on an already-`Funded` intent.

**Options 2 and 3 do not fix the stated impact.** Both close only the `openFor` door. An attacker wanting duplicate `Open` events would simply call `open()` — no signature, no deadline, unbounded in time. Shipping either would close a keyhole beside an open door *while retiring the ticket*, which is worse than doing nothing.

A complete on-chain fix means making `publish` revert on an existing intent, which breaks behavior that is deliberate and useful: announcing an intent escrowed via `fund`/`fundFor`, and replaying an announcement missed during an indexer outage or reorg. It would also turn a harmless solver race into a revert.

`Portal` is not a proxy, so any bytecode change means a fresh CREATE2 deployment across 15+ chains, a version bump, and integrator migration — a large operation to buy a property the `publish`/`open` paths still would not have.

Since the exposure is off-chain, the remedy is off-chain idempotency. What makes that remedy *reachable* is accurate documentation, so the misleading NatSpec is treated as the actual defect.

## Boundaries

- **No contract behavior change.** Deliberate, per the above. Confirmed by byte-identical deployed bytecode.
- **Exactly-once events are not delivered.** If they are ever a hard requirement, the change belongs at `publish`, not at the gasless nonce, and it is a breaking behavior change for a deliberate major version — not a patch.
- **The nonce still does nothing.** Left as-is; making it meaningful is the same major-version design question.
- **Prettier:** all files changed here pass `prettier --check`. The only `contracts/` edits are comments, so no executable line changes and `Portal` bytecode is unchanged. (A pre-existing #421 collapse of `revert InvalidCreatorBinding(...)` at printWidth 80 exists on `origin/main`; CI runs `yarn format` but does not gate on the diff.)

## Validation

| Gate | Result |
| --- | --- |
| `forge build --force` | Compiler run successful; warnings only, all pre-existing |
| `forge test --match-contract OriginSettlerTest` | **17 passed, 0 failed, 0 skipped** |
| `forge test` (full) | **604 passed, 0 failed, 0 skipped** (32 suites) |
| `yarn test:hardhat` | 112 passing |
| `yarn test:ts` | 9 passed |
| Deployed-bytecode diff (`Portal`) | **Byte-identical** before/after — `sha256 a4544c3d…371b` |
| Non-comment lines changed in `contracts/` | **Zero** (verified by filtering the diff) |

On test shape: because nothing about the behavior changes, these are **characterization** tests that pass against unmodified `origin/main` — not failing-first tests. They were run against pristine `origin/main` before any comment was touched, and all seven passed; that run is the repro. Writing them to fail would have meant asserting behavior that was then deliberately not implemented.

- `testOpenForReplayEmitsDuplicateOpenAndMovesNoFunds` — over-mints and over-approves first, so "no second pull" is a real assertion rather than an artifact of an exhausted balance.
- `testOpenForDistinctNoncesCollapseToOneOrderIdAndOneEscrow` — now asserts exactly one `Open` per call and anchors both orderIds to the derived `intentHash` (previously compared orderIds only).
- `testPublishIsPermissionlesslyRepeatableOnFundedIntent`
- `testOpenIsPermissionlesslyRepeatableOnFundedIntent`
- `testTerminalStateStopsReannouncement` — `Withdrawn`/`Refunded` stops re-announcement (`IntentAlreadyExists`).
- `testOpenForReplayReemitsIntentPublished` — pins the `IntentPublished` re-emission on the `openFor` path.
- `testOpenForReplayWithNativeRefundsCallerAndLeavesVaultUntouched` — native `openFor{value}` replay refunds the caller, vault untouched.

`yarn lint` / `yarn format` were **not** run: `format:solhint` uses `--fix`, which is known in this repo to rename underscore-named contracts and break the build. `prettier --check` was run on the changed files instead.

Base is `origin/main` @ `c8603d01`. #421's `order.user == reward.creator` check was reviewed and does not affect replay — it constrains *who* can be named, not how many times a signature works.


MSevey (commented):
Reviewed the docs against the source and reproduced the safety claim. One real defect, three smaller notes.

### Verified

Rebuilt `Portal` at base `c8603d01` and at head `c1054e7f`:

```
BASE deployedBytecode sha256: a4544c3d2b45971c1ab5d6317caa7a9b3e549aada617b59e8138339e7640371b
HEAD deployedBytecode sha256: a4544c3d2b45971c1ab5d6317caa7a9b3e549aada617b59e8138339e7640371b
```

Identical, and matches the hash in the description. All 14 tests in `OriginSettlerTest` pass. Test file is purely additive (zero deletions).

Also spot-checked and confirmed: `onlyFundable` returns on `Funded` (`IntentSource.sol:84`), partial funding leaves `Initial` (`if (fullyFunded)`, `:673`), `_validatePublish` rejects only terminal states, `orderId == intentHash` with `nonce` excluded, `nonce` present in the EIP-712 typehash but never consumed, `openDeadline` enforced in `openFor`, and `otherPerson` genuinely holds no reward tokens and no approval (so that comment in `testOpenIsPermissionlesslyRepeatableOnFundedIntent` is accurate).

---

### 1. The `Open` NatSpec names two functions that never emit `Open`

`contracts/interfaces/ERC7683/IOriginSettler.sol` — cause 1 reads:

> any further `open`/`openFor`/`publish`/`publishAndFund` for the same intent re-emits the announcement … `open` and `publish` take no signature and no deadline

`Open` is emitted in exactly two places:

```
contracts/ERC7683/OriginSettler.sol:70   emit Open(...)   // open()
contracts/ERC7683/OriginSettler.sol:142  emit Open(...)   // openFor()
```

`publish()` and `publishAndFund()` emit `IntentPublished`, never `Open`.

The mirror error is in the same diff: `IIntentSource.sol` documents `IntentPublished` as reachable only via `publish`, but `_publishAndFund` calls `publish()` internally (`IntentSource.sol:626`), so `open`, `openFor`, and `publishAndFund` all re-emit `IntentPublished` too. The two enumerations read as swapped — `Open` over-lists by two, `IntentPublished` under-lists by three.

Worth fixing precisely because it is the failure mode this PR exists to correct. An ERC-7683 integrator reading the `Open` NatSpec is told to expect duplicates from `publish` (they never arrive), and reading the `IntentPublished` NatSpec is not told that `open`/`openFor` duplicate it (they do). The at-least-once conclusion is unaffected — `open()` alone carries it — but the causal map is wrong.

Suggested wording:

```solidity
// IOriginSettler.sol, cause 1
//   1. Publishing is intentionally idempotent. `orderId` is the intent hash,
//      and once its rewards are escrowed (`Status.Funded`) a further `open` or
//      `openFor` for the same intent re-emits `Open` and leaves the escrow
//      untouched. `open` takes no signature and no deadline, so this is
//      permissionless until the intent reaches a terminal state. Note that
//      `publish`/`publishAndFund` do NOT emit `Open`; they re-emit
//      `IntentPublished` -- see {IIntentSource}.

// IIntentSource.sol
//      `publish` is permissionless, takes no deadline, and only rejects
//      intents already `Withdrawn`/`Refunded`. `publishAndFund`, `open`, and
//      `openFor` each call `publish` internally, so all four re-emit this event.
```

### 2. `README.md` — same conflation, softer

> any further `open`, `openFor`, `publish`, or `publishAndFund` for that same intent re-emits the announcement

Defensible since the section covers both events jointly, but "the announcement" invites the same misread. Naming which event each entry point emits would close it.

### 3. "unbounded in time" overstates the window

`IOriginSettler.sol` says "permissionless and unbounded in time"; the README says "at any time". Re-announcement actually stops at the terminal states:

```
contracts/IntentSource.sol:473   rewardStatuses[intentHash] = Status.Withdrawn;
contracts/IntentSource.sol:950   rewardStatuses[intentHash] = Status.Refunded;
```

after which `_validatePublish` reverts `IntentAlreadyExists`. The terminal rule is stated in a different paragraph, so a careful reader reconciles it — but "until withdrawn or refunded" is more accurate and more useful to an indexer deciding how long to keep deduping.

### 4. Three test coverage gaps

- **No test that a terminal state stops re-announcement.** This is the bound on the whole "permissionless forever" claim, and the invariant most worth pinning: if a future change made `Refunded` re-publishable, the docs go silently wrong *and* it is a real problem. `refund` → `publish` → `expectRevert(IntentAlreadyExists)`.
- **No assertion that the replay re-emits `IntentPublished`.** `_countOpenLogs` filters on `Open.selector` only, so the claim the `IntentPublished` NatSpec makes is untested on the `openFor` path.
- **Native rewards never exercised.** `BaseTest` sets `reward.nativeAmount: 0`, so a replayed `openFor{value: X}` on a `Funded` intent is untested. That path takes the `onlyFundable` early return and then `Refund.excessNative()` (`IntentSource.sol:628`) refunds to `msg.sender`. Worth pinning that the replayer gets their ETH back and the vault is untouched.

---

Related check that came up and cleared: `Refund.excessNative()` sends `address(this).balance` rather than `msg.value`, and `open()` is a permissionless no-op on a `Funded` intent. Not exploitable — `Portal` has no `receive()`/`fallback()` (only `Executor`, `LocalProver`, and `vault/Proxy` do) and every payable entry point refunds within the same transaction, so `Portal` holds no balance at rest.

Only item 1 seems worth blocking on.

carlosfebres:
Addressed all four in `ad5d74fe`, @MSevey.

**1 (blocking) — swapped enumerations.** Fixed on both interfaces. The `Open` NatSpec now states only `open`/`openFor` emit it and points to {IIntentSource} for `IntentPublished`; the `IntentPublished` NatSpec now spells out that `publishAndFund`, `open`, and `openFor` each call `publish` internally, so all four re-emit it.

**2 — README conflation.** The section now names which event each entry point emits instead of "the announcement".

**3 — "unbounded in time".** Reworded to "until the intent reaches a terminal state (`Withdrawn`/`Refunded`), after which publishing reverts" in both the NatSpec and the README.

**4 — test gaps.** Added three:
- `testTerminalStateStopsReannouncement` — refund → publish → `IntentAlreadyExists`.
- `testOpenForReplayReemitsIntentPublished` — pins the mirror claim on the `openFor` path, which `_countOpenLogs` (Open-only) couldn't see.
- `testOpenForReplayWithNativeRefundsCallerAndLeavesVaultUntouched` — the native path, since BaseTest uses `nativeAmount == 0`; asserts the replayer's forwarded ETH is refunded and the vault is untouched.

604 tests pass, prettier clean.

contracts/interfaces/IIntentSource.sol
```diff
@@ -72,6 +72,23 @@ interface IIntentSource {
...
+     *      intents that are already `Withdrawn` or `Refunded`. Re-publishing an
+     *      `Initial` or `Funded` intent therefore succeeds and re-emits this
+     *      event; when the intent is already `Funded` the escrow is left
+     *      untouched, so the repeat is free to trigger and moves no funds. This
+     *      is deliberate -- it lets an intent escrowed via `fund`/`fundFor` be
+     *      announced after the fact, and lets a missed announcement be replayed
+     *      for indexers.
+     *
+     *      `publishAndFund`, `open`, and `openFor` each call `publish`
+     *      internally, so all four entry points re-emit this event -- not just
```
MSevey: **Incomplete: `publishAndFundFor` also publishes.** `contracts/IntentSource.sol:419-427`:

```solidity
function publishAndFundFor(
    uint64 destination, bytes memory route, Reward calldata reward,
    bool allowPartial, address funder, address permitContract
) public payable returns (bytes32 intentHash, address vault) {
    (intentHash, ) = publish(destination, route, reward);   // ← emits IntentPublished
```

Both overloads (`:391`, `:419`) route through `publish`, so it's five entry points, not four. Worth stating explicitly that `fund`/`fundFor` escrow *without* publishing so they emit neither — that's the other half an integrator needs.

While here: a re-emission is never accompanied by `IntentFunded` (the `onlyFundable` early return skips that emit), which consumers pairing the two events need to know.

README.md
```diff
@@ -122,6 +122,19 @@ Eco Routes implements [ERC-7683](https://eips.ethereum.org/EIPS/eip-7683), the s
...
 
+#### Event Semantics for Integrators
+
+`Open` and `IntentPublished` are **at-least-once, not exactly-once**. Indexers, solvers, and quoting services **must dedupe on `orderId` / `intentHash`** and must not treat each event as a distinct order.
+
+Publishing is intentionally idempotent. `orderId` equals the intent hash — a pure function of `(destination, routeHash, rewardHash)` — and once an intent's rewards are escrowed, re-publishing that same intent re-emits its announcement and leaves the escrow untouched. The two events come from different entry points: `Open` is emitted only by `open` and `openFor`, while `IntentPublished` is emitted by `publish`, `publishAndFund`, `open`, and `openFor` (the last three call `publish` internally). `open` and `publish` require no signature and impose no deadline, so anyone can re-announce a funded intent for the cost of gas — until it reaches a terminal state (`Withdrawn` or `Refunded`), after which publishing reverts. This is a deliberate property: it lets an intent escrowed via `fund`/`fundFor` be announced afterwards, and lets an announcement missed during an outage or reorg be replayed for indexers.
```
MSevey: Same gap as the NatSpec: `publishAndFundFor` is missing from the `IntentPublished` list, so this reads as four entry points when there are five. Suggest also noting that `fund`/`fundFor` escrow without publishing and therefore emit neither.

README.md
```diff
@@ -122,6 +122,19 @@ Eco Routes implements [ERC-7683](https://eips.ethereum.org/EIPS/eip-7683), the s
...
+
+Two consequences worth calling out:
+
+- **A gasless signature is replayable until `openDeadline`.** `GaslessCrossChainOrder.nonce` is covered by the EIP-712 digest but is not consumed on-chain, so resubmitting the same `openFor` payload succeeds and emits `Open` again.
```
MSevey: Missing the terminal-state bound that the `Open` NatSpec in `IOriginSettler` carries: replay stops at `openDeadline` **or** when the intent hits `Withdrawn`/`Refunded`, whichever comes first. `testTerminalStateStopsReannouncement` pins the tighter bound, so the docs should match it.

README.md
```diff
@@ -122,6 +122,19 @@ Eco Routes implements [ERC-7683](https://eips.ethereum.org/EIPS/eip-7683), the s
...
+- **`nonce` does not distinguish orders.** Two signed gasless orders differing only by `nonce` resolve to the same `orderId` and the same single escrow. To create genuinely distinct intents, vary a field that feeds the hash — normally `route.salt`.
+
+Duplicate announcements never move funds twice (escrow is single-shot) and duplicate fills are rejected on the destination chain, so the risk is confined to off-chain consumers that count events rather than deduping them. A consumer that double-counts can double-quote or attempt a fill it cannot win.
```
MSevey, line 138: Worth adding here: a duplicate announcement arrives **without** a matching `IntentFunded`. The `onlyFundable` early return on `Status.Funded` skips the funding body and its `emit IntentFunded` (`contracts/IntentSource.sol:677`, `:772`), so an indexer that correlates the two events sees an orphan `IntentPublished`. That's exactly the class of surprise this section exists to remove.

contracts/IntentSource.sol
```diff
@@ -62,7 +62,16 @@ abstract contract IntentSource is OriginSettler, IIntentSource {
     /**
      * @notice Ensures intent can be funded based on its current status
      * @dev Prevents funding of intents that are already withdrawn or refunded
-     *      Allows funding of Initial or Funded status intents (for partial funding)
+     *      Allows funding of Initial status intents (including top-ups after
+     *      partial funding, which leaves the status Initial)
+     * @dev On `Status.Funded` this RETURNS rather than reverting, skipping the
+     *      decorated body entirely. That makes escrow single-shot: a repeated
+     *      fund/publishAndFund/open/openFor for an already-escrowed intent is a
```
MSevey: The entry-point list is missing `fundFor` and `publishAndFundFor`. `_fundIntentFor` carries the same modifier — `contracts/IntentSource.sol:752`:

```solidity
) internal onlyFundable(intentHash) returns (address vault) {
```

So those two paths take the same early return, and for them the *named `funder`* is silently never charged — arguably the more surprising consequence than the `msg.sender` case.

Also worth stating: the early return skips the body's `emit IntentFunded`, so a no-op call emits no funding event at all.

contracts/ERC7683/OriginSettler.sol
```diff
@@ -69,9 +74,14 @@ abstract contract OriginSettler is IOriginSettler, EIP712 {
      * @notice Opens an Eco intent on behalf of a user via ERC-7683 gasless interface
      * @dev Called by a solver to create an intent for a user using their signature
      * @dev Performs comprehensive validation: deadlines, signature, chain IDs, origin settler
-     * @dev Includes replay protection through vault state checking in _publishAndFund
      * @dev Uses unified _publishAndFund method for consistent behavior and security
      * @dev Emits Open event with ERC-7683 compliant ResolvedCrossChainOrder
+     * @dev `order.nonce` is covered by the EIP-712 digest but is NOT consumed
+     *      on-chain, so a signature remains usable until `openDeadline` passes.
```
MSevey: "until `openDeadline` passes" is the looser of the two bounds. Publishing also stops once the intent reaches `Withdrawn`/`Refunded` — which the `Open` NatSpec in `IOriginSettler` states and `testTerminalStateStopsReannouncement` proves. Three doc sites, two of which omit it; worth aligning.

test/interfaces/OriginSettler.t.sol
```diff
@@ -222,6 +218,436 @@ contract OriginSettlerTest is BaseTest {
...
+            keccak256(sigTwo),
+            "differing nonces must produce differing signatures"
+        );
+
+        address vault = intentSource.intentVaultAddress(intent);
+
+        vm.recordLogs();
+        vm.prank(otherPerson);
+        portal.openFor(orderOne, sigOne, "");
+        (, bytes32 orderIdOne) = _countOpenLogs();
```
MSevey: **This assertion can pass vacuously.** The count is discarded, so if zero `Open` records match, `orderIdOne` and `orderIdTwo` are both `bytes32(0)` and the comparison below succeeds.

Confirmed by mutation — replacing `openSig` with `keccak256("NeverMatches()")`:

```
[PASS] testOpenForDistinctNoncesCollapseToOneOrderIdAndOneEscrow
[FAIL: first openFor should emit one Open: 0 != 1] testOpenForReplayEmitsDuplicateOpenAndMovesNoFunds
[FAIL: unsigned open re-emits Open on a funded intent: 0 != 1] testOpenIsPermissionlesslyRepeatableOnFundedIntent
```

The two siblings catch the regression because they assert the count; this one doesn't. Fix is what `testOpenForReplayEmitsDuplicateOpenAndMovesNoFunds` already does — assert `count == 1` and anchor `orderIdOne` against the derived `intentHash`. With that, the same mutation fails here too. Patch is in the review summary.

MSevey (commented):
## Review — verified the claims, found one real gap in the emit map

Reviewed at `ad5d74fe` against `origin/main@c8603d01`, built and tested in an isolated worktree.

**All three headline claims hold:**

| Claim | Verdict |
| --- | --- |
| `Portal` deployed bytecode byte-identical | ✅ `a4544c3d…371b` on both base and head |
| Zero non-comment lines changed in `contracts/` | ✅ filtered the diff, empty |
| `forge test --match-contract OriginSettlerTest` | ✅ **17 passed, 0 failed** |

The reasoning in "Why This Shape" is right: options 2 and 3 close `openFor` while `open()` stays wide open, so documenting the at-least-once contract is the correct remedy. Six issues below, one of which matters.

### 1. The emit map is still incomplete — `publishAndFundFor` is a fifth entry point (P1)

`publishAndFundFor` publishes, so it re-emits `IntentPublished`, but it is missing from both the NatSpec and the README. Since the entire deliverable here is an accurate event map, an integrator reading "these four entry points" will watch four and miss the fifth — the exact failure this PR exists to prevent. `ad5d74fe` corrected the map for `publish`/`publishAndFund` but not this one. Inline comments below.

### 2. `onlyFundable`'s entry-point list omits `fundFor` / `publishAndFundFor`

`_fundIntentFor` carries the same modifier, so those paths take the same early return — and for them the *named `funder`* is silently never charged, which is the more surprising half.

### 3. The early return also skips `IntentFunded` — undocumented

`emit IntentFunded` sits inside the modifier-guarded body, so a duplicate announcement arrives with **no matching `IntentFunded`**. An indexer correlating the pair sees an orphan `IntentPublished`. That belongs in an "Event Semantics for Integrators" section and is currently absent from all four doc sites.

### 4. One of the new characterization tests passes vacuously

`testOpenForDistinctNoncesCollapseToOneOrderIdAndOneEscrow` discards the log count and compares only the two orderIds, so `bytes32(0) == bytes32(0)` satisfies it when zero `Open` events are observed. Proven by mutation — replacing the topic filter with `keccak256("NeverMatches()")`:

```
[PASS] testOpenForDistinctNoncesCollapseToOneOrderIdAndOneEscrow
[FAIL: first openFor should emit one Open: 0 != 1] testOpenForReplayEmitsDuplicateOpenAndMovesNoFunds
[FAIL: unsigned open re-emits Open on a funded intent: 0 != 1] testOpenIsPermissionlesslyRepeatableOnFundedIntent
```

Its two siblings catch it; this one doesn't. For a test whose only job is to fail loudly on a future behavior change, that's worth fixing. After the patch below, the same mutation fails it.

### 5. "usable until `openDeadline`" is missing the terminal-state bound

Two of the three sites say `openDeadline`; the `Open` NatSpec in `IOriginSettler` correctly adds the terminal state. `testTerminalStateStopsReannouncement` proves the tighter bound.

### 6. PR description is stale by one commit

The Validation table says 4 characterization tests / 14 passing; head has 7 new tests / 17 passing (`ad5d74fe` added terminal-state, `IntentPublished` mirror, native-replay). Also "That reformatting was **reverted**" is backwards for the test file: `test/interfaces/OriginSettler.t.sol:200` still carries the prettier collapse — I confirmed both files are prettier-dirty on base and only the contract's drift was reverted. On a PR whose deliverable is a trustworthy written record, the body should match the diff.

---

### Ready-to-apply patch

Verified in an isolated worktree: **17/17 pass**, `Portal` bytecode **still `a4544c3d…371b`** (every contract edit is a comment), prettier clean on all touched files except the pre-existing `OriginSettler.sol` drift already noted in the description.

<details>
<summary><code>pr426-review-fixes.patch</code> — 5 files, +49/-15</summary>

```diff
diff --git a/README.md b/README.md
index 8060c78c..b82ff25f 100644
--- a/README.md
+++ b/README.md
@@ -126,13 +126,15 @@ Users can create and fulfill intents through both our native interface and the s
 
 `Open` and `IntentPublished` are **at-least-once, not exactly-once**. Indexers, solvers, and quoting services **must dedupe on `orderId` / `intentHash`** and must not treat each event as a distinct order.
 
-Publishing is intentionally idempotent. `orderId` equals the intent hash — a pure function of `(destination, routeHash, rewardHash)` — and once an intent's rewards are escrowed, re-publishing that same intent re-emits its announcement and leaves the escrow untouched. The two events come from different entry points: `Open` is emitted only by `open` and `openFor`, while `IntentPublished` is emitted by `publish`, `publishAndFund`, `open`, and `openFor` (the last three call `publish` internally). `open` and `publish` require no signature and impose no deadline, so anyone can re-announce a funded intent for the cost of gas — until it reaches a terminal state (`Withdrawn` or `Refunded`), after which publishing reverts. This is a deliberate property: it lets an intent escrowed via `fund`/`fundFor` be announced afterwards, and lets an announcement missed during an outage or reorg be replayed for indexers.
+Publishing is intentionally idempotent. `orderId` equals the intent hash — a pure function of `(destination, routeHash, rewardHash)` — and once an intent's rewards are escrowed, re-publishing that same intent re-emits its announcement and leaves the escrow untouched. The two events come from different entry points: `Open` is emitted only by `open` and `openFor`, while `IntentPublished` is emitted by `publish`, `publishAndFund`, `publishAndFundFor`, `open`, and `openFor` (the last four call `publish` internally). `fund` and `fundFor` escrow without publishing, so they emit neither. `open` and `publish` require no signature and impose no deadline, so anyone can re-announce a funded intent for the cost of gas — until it reaches a terminal state (`Withdrawn` or `Refunded`), after which publishing reverts. This is a deliberate property: it lets an intent escrowed via `fund`/`fundFor` be announced afterwards, and lets an announcement missed during an outage or reorg be replayed for indexers.
 
 Two consequences worth calling out:
 
-- **A gasless signature is replayable until `openDeadline`.** `GaslessCrossChainOrder.nonce` is covered by the EIP-712 digest but is not consumed on-chain, so resubmitting the same `openFor` payload succeeds and emits `Open` again.
+- **A gasless signature is replayable until `openDeadline`** (or until the intent reaches a terminal state, whichever comes first). `GaslessCrossChainOrder.nonce` is covered by the EIP-712 digest but is not consumed on-chain, so resubmitting the same `openFor` payload succeeds and emits `Open` again.
 - **`nonce` does not distinguish orders.** Two signed gasless orders differing only by `nonce` resolve to the same `orderId` and the same single escrow. To create genuinely distinct intents, vary a field that feeds the hash — normally `route.salt`.
 
+A duplicate announcement also arrives **without** a matching `IntentFunded`, because the funding path returns early on an already-escrowed intent and never reaches that emit. Consumers that correlate the two events must tolerate the gap.
+
 Duplicate announcements never move funds twice (escrow is single-shot) and duplicate fills are rejected on the destination chain, so the risk is confined to off-chain consumers that count events rather than deduping them. A consumer that double-counts can double-quote or attempt a fill it cannot win.
 
 ## Setup & Installation
diff --git a/contracts/ERC7683/OriginSettler.sol b/contracts/ERC7683/OriginSettler.sol
index a05b009b..d1aeffff 100644
--- a/contracts/ERC7683/OriginSettler.sol
+++ b/contracts/ERC7683/OriginSettler.sol
@@ -77,7 +77,9 @@ abstract contract OriginSettler is IOriginSettler, EIP712 {
      * @dev Uses unified _publishAndFund method for consistent behavior and security
      * @dev Emits Open event with ERC-7683 compliant ResolvedCrossChainOrder
      * @dev `order.nonce` is covered by the EIP-712 digest but is NOT consumed
-     *      on-chain, so a signature remains usable until `openDeadline` passes.
+     *      on-chain, so a signature remains usable until `openDeadline` passes
+     *      or the intent reaches a terminal state (`Withdrawn`/`Refunded`),
+     *      whichever comes first.
      *      Escrow is still single-shot (see the vault state check in
      *      `_publishAndFund`), but a replay does re-emit `Open`. `order.nonce`
      *      also does not feed `orderId`, so two orders differing only by nonce
diff --git a/contracts/IntentSource.sol b/contracts/IntentSource.sol
index b4f47868..b1164c2a 100644
--- a/contracts/IntentSource.sol
+++ b/contracts/IntentSource.sol
@@ -66,12 +66,18 @@ abstract contract IntentSource is OriginSettler, IIntentSource {
      *      partial funding, which leaves the status Initial)
      * @dev On `Status.Funded` this RETURNS rather than reverting, skipping the
      *      decorated body entirely. That makes escrow single-shot: a repeated
-     *      fund/publishAndFund/open/openFor for an already-escrowed intent is a
-     *      successful no-op instead of a second charge, and the caller needs no
-     *      balance or allowance to make that call. Callers that need to know
-     *      whether they were the one to escrow the intent must compare
-     *      {getRewardStatus} before and after; a successful call alone does not
-     *      imply funds moved.
+     *      call for an already-escrowed intent is a successful no-op instead of
+     *      a second charge, and the caller needs no balance or allowance to make
+     *      it. This applies to every funding entry point, because both
+     *      {_fundIntent} and {_fundIntentFor} carry this modifier -- so `fund`,
+     *      `fundFor`, `publishAndFund`, `publishAndFundFor`, `open`, and
+     *      `openFor` all take the early return. For the `*For` variants the
+     *      named `funder` is likewise never charged.
+     * @dev The early return also skips the body's `IntentFunded` emit, so a
+     *      no-op call emits no funding event (a re-publishing entry point still
+     *      emits `IntentPublished`). Callers that need to know whether they were
+     *      the one to escrow the intent must compare {getRewardStatus} before
+     *      and after; a successful call alone does not imply funds moved.
      * @param intentHash Hash of the intent to validate for funding eligibility
      */
     modifier onlyFundable(bytes32 intentHash) {
diff --git a/contracts/interfaces/IIntentSource.sol b/contracts/interfaces/IIntentSource.sol
index 9dee3266..4eb77a2b 100644
--- a/contracts/interfaces/IIntentSource.sol
+++ b/contracts/interfaces/IIntentSource.sol
@@ -85,10 +85,17 @@ interface IIntentSource {
      *      announced after the fact, and lets a missed announcement be replayed
      *      for indexers.
      *
-     *      `publishAndFund`, `open`, and `openFor` each call `publish`
-     *      internally, so all four entry points re-emit this event -- not just
-     *      `publish`. (`Open` is separate and is emitted only by `open` and
-     *      `openFor`; see {IOriginSettler}.)
+     *      `publishAndFund`, `publishAndFundFor`, `open`, and `openFor` each
+     *      call `publish` internally, so all five entry points re-emit this
+     *      event -- not just `publish`. `fund` and `fundFor` do NOT publish, so
+     *      they never emit this event. (`Open` is separate and is emitted only
+     *      by `open` and `openFor`; see {IOriginSettler}.)
+     *
+     *      A re-emission is not accompanied by `IntentFunded`: the
+     *      {IntentSource-onlyFundable} check returns early on `Status.Funded`,
+     *      skipping the funding body and its `IntentFunded` emit. Consumers
+     *      that pair the two events must tolerate an `IntentPublished` with no
+     *      matching `IntentFunded`.
      * @param intentHash Unique identifier of the intent
      * @param destination Destination chain ID
      * @param route Encoded route data for the destination chain
diff --git a/test/interfaces/OriginSettler.t.sol b/test/interfaces/OriginSettler.t.sol
index 41ff01b4..da042e06 100644
--- a/test/interfaces/OriginSettler.t.sol
+++ b/test/interfaces/OriginSettler.t.sol
@@ -412,10 +412,26 @@ contract OriginSettlerTest is BaseTest {
 
         address vault = intentSource.intentVaultAddress(intent);
 
+        (bytes32 intentHash, , ) = intentSource.getIntentHash(
+            CHAIN_ID,
+            abi.encode(route),
+            reward
+        );
+
         vm.recordLogs();
         vm.prank(otherPerson);
         portal.openFor(orderOne, sigOne, "");
-        (, bytes32 orderIdOne) = _countOpenLogs();
+        (uint256 countOne, bytes32 orderIdOne) = _countOpenLogs();
+
+        // Anchor on a non-zero observation, so a future change that stops
+        // emitting `Open` fails here instead of letting the orderId comparison
+        // below pass vacuously on bytes32(0) == bytes32(0).
+        assertEq(countOne, 1, "first openFor should emit one Open");
+        assertEq(
+            orderIdOne,
+            intentHash,
+            "orderId must equal the derived intentHash"
+        );
 
         uint256 vaultAAfterFirst = tokenA.balanceOf(vault);
         uint256 vaultBAfterFirst = tokenB.balanceOf(vault);
@@ -423,8 +439,9 @@ contract OriginSettlerTest is BaseTest {
         vm.recordLogs();
         vm.prank(otherPerson);
         portal.openFor(orderTwo, sigTwo, "");
-        (, bytes32 orderIdTwo) = _countOpenLogs();
+        (uint256 countTwo, bytes32 orderIdTwo) = _countOpenLogs();
 
+        assertEq(countTwo, 1, "second nonce should emit one Open");
         assertEq(
             orderIdTwo,
             orderIdOne,
```

</details>

fish-sammy (dismissed):

carlosfebres:
Addressed your 2nd-round items in `3e6722aa`, @MSevey:

**1 (the one that matters)** — added `publishAndFundFor` as the 5th entry point that re-emits `IntentPublished` (verified it calls `publish()` at `IntentSource.sol:427`), on both the NatSpec and README. Also noted `fund`/`fundFor` escrow WITHOUT publishing, so they never emit it.
**2** — the `onlyFundable` NatSpec now lists `fundFor`/`publishAndFundFor`, and calls out that the `*For` variants silently don't charge the named funder on the early return.
**3** — documented that the early return also skips `IntentFunded`, so a replayed `IntentPublished` on a Funded intent is an orphan for pair-correlating indexers (NatSpec + README).
**4** — fixed the vacuous `testOpenForDistinctNoncesCollapseToOneOrderIdAndOneEscrow`: it now asserts exactly one `Open` per call and anchors both orderIds to the derived `intentHash`. Mutation-checked — removing the first `openFor` now fails it (`0 != 1`).
**5** — brought `OriginSettler.sol:80` + README into line with the `IOriginSettler` NatSpec: replay is bounded by min(`openDeadline`, terminal `Withdrawn`/`Refunded`).
**6** — refreshed the description (7 new tests / 604 total; corrected the "reformatting reverted" wording).

Full forge suite 604 pass, prettier clean, `Portal` bytecode unchanged (comments/tests only).

contracts/interfaces/IIntentSource.sol
carlosfebres: Added `publishAndFundFor` as the fifth entry point that re-emits `IntentPublished` (verified it calls `publish()` at IntentSource.sol:427), and noted `fund`/`fundFor` escrow without publishing so they never emit it. 3e6722aa.

README.md
carlosfebres: README now lists all five entry points including `publishAndFundFor`, plus the `fund`/`fundFor` no-publish note. 3e6722aa.

README.md
carlosfebres: Added the terminal-state bound to the README: replay is usable until min(`openDeadline`, terminal `Withdrawn`/`Refunded`). 3e6722aa.

README.md
carlosfebres: Documented it: the early return skips `IntentFunded`, so a replayed `IntentPublished` on a Funded intent is an orphan for pair-correlating indexers. 3e6722aa.

contracts/IntentSource.sol
carlosfebres: The `onlyFundable` NatSpec now lists `fundFor`/`publishAndFundFor`, and calls out that the `*For` variants silently do not charge the named funder on the early return. 3e6722aa.

contracts/ERC7683/OriginSettler.sol
carlosfebres: Brought OriginSettler.sol:80 into line with the `IOriginSettler` `Open` NatSpec — replay is bounded by min(`openDeadline`, terminal `Withdrawn`/`Refunded`). 3e6722aa.

test/interfaces/OriginSettler.t.sol
carlosfebres: Fixed. The test now asserts exactly one `Open` per call and anchors both orderIds to the derived `intentHash`. Mutation-checked: removing the first `openFor` now fails it (`0 != 1`). 3e6722aa.

MSevey (approved):
Approving. Verified the central claim independently rather than taking the PR body's word for it:

- **Zero non-comment lines changed in `contracts/`** — filtered every added/removed line through a comment stripper; nothing survived.
- **`Portal` deployed bytecode byte-identical** — built merge-base `e5d4ad47` and PR head separately, `sha256 4e09e474e864…14df` on both sides. `bytecode_hash = "none"` + `cbor_metadata = false` in `foundry.toml` means that's a real comparison, not metadata noise.
- **Test diff purely additive** — no deleted or modified lines, no `skip`/`.only`/`xit`.
- `forge test --match-contract OriginSettlerTest` → **27 passed, 0 failed** locally.

Since the deliverable is integrator-facing documentation on a fund-critical contract, I checked each substantive claim against source instead of reviewing the prose. All hold: `onlyFundable` returns (not reverts) on `Funded` (`IntentSource.sol:87-89`); `_validatePublish` rejects only `Withdrawn`/`Refunded` (`:848`); `intentHash = keccak256(destination, routeHash, rewardHash)` with no nonce (`:172`); `open()` checks only `orderDataType`, no signature and no deadline; `Open` is emitted only at `OriginSettler.sol:69,142`; both `_publishAndFund` and `publishAndFundFor` route through `publish`, so the "five entry points" claim is right; `emit IntentFunded` sits behind `onlyFundable`, so the orphan-`IntentPublished` claim is right; partial funding leaves status `Initial` because `Funded` is set only `if (fullyFunded)`.

The reasoning for not fixing this on-chain is sound and worth keeping in the record — consuming the nonce or reverting `openFor` on `Funded` closes only the `openFor` door while `open()` stays permissionless and unbounded, i.e. closing a keyhole beside an open door while retiring the ticket.

Two non-blocking observations, neither of which should hold the merge:

1. **`contracts/interfaces/ERC7683/IOriginSettler.sol:36`** — the docs say to dedupe on `orderId` but don't mention that duplicate payloads are not identical: `Open.resolvedOrder.openDeadline` is `uint32(block.timestamp)` via `open()` but `order.openDeadline` via `openFor()`. A first-write-wins indexer and a last-write-wins indexer will therefore record different deadlines for the same `orderId`. Nothing written is false, so this is an omission rather than an error, but it's about one sentence from complete and it's the same class of consumer bug this PR is closing.

2. **`contracts/IntentSource.sol:70-75`** — the new `onlyFundable` NatSpec adds the genuinely surprising claim that for `fundFor`/`publishAndFundFor` the named `funder` is silently *not* charged on the early return, and no characterization test pins it. The closest existing test, `testFundForAlreadyFundedDoesNotStrandNative` (`test/source/IntentSource.t.sol:583`), uses `funder == caller` and asserts only the native leg. Every other claim in this PR got a test; this is the one most likely to bite a relayer author who reads "call succeeded" as "funder was charged". The higher-value of the two if you want either landed.

Nit: the validation table says `OriginSettlerTest` → 17 passed; it's 27 now, stale after the `origin/main` merge in `e6090586`.

PR #433, opened 2026-08-20
feat(prover): add AggregatorProver, a 1-of-N union prover

carlosfebres:
## Problem

`reward.prover` is a single address, fixed by the intent **creator** at publish time
(`Reward.prover`, baked into the intent hash). Which bridge is actually live, uncensored, and
affordable is only knowable by the **solver** at fulfillment time.

A creator who names `HyperProver` and then finds Hyperlane degraded has no recourse: the solver
either proves over that one named bridge or does not get paid. The multi-prover architecture already
exists — six prover implementations — but an intent can only ever point at one of them.

## Impact

Latent liveness and settlement risk, not a defect in any deployed contract.

- **Worst case:** a solver fulfills on the destination chain, the creator-named bridge is down or
  censoring, no proof lands before `reward.deadline`, and the creator refunds. The solver delivered
  goods and is not paid.
- **Conditions:** one bridge degrades between publish and fulfillment. No compromise required.
- **Active or latent:** latent. Nothing is broken today; the protocol simply cannot express
  "any of these bridges is acceptable."

## What Changed

- **`contracts/prover/AggregatorProver.sol`** (new, 315 lines) — implements `IProver` directly. Stateless,
  read-only. Reports an intent proven when **any one** member of an immutable member set has proven
  it (1-of-N union).
  - `provenIntents(bytes32)` — fans out read-only calls to members in immutable priority order,
    returns the first proof with a non-zero claimant.
  - `challengeIntentProof(uint64,bytes32,bytes32)` — forwards to every member in `try`/`catch`;
    never reverts.
  - `prove(...)` — reverts `ProvingNotSupported`. The contract dispatches no messages; solvers prove
    through a concrete member via `Portal.prove`.
  - `getMembers()`, `getProofType()` (`"Aggregator"`), `supportsInterface`.
  - Membership is constructor-set and immutable, capped at `MAX_MEMBERS = 8`, validated for
    non-zero, EVM-shaped, and non-duplicate. **No owner, no setter, no admin role.**
- **`scripts/Deploy.s.sol`** — deploys `AggregatorProver` last (its members must already exist), from the
  ordered `AGGREGATOR_PROVER_MEMBERS` env var; `validateAggregatorProverMembers` gates membership at deploy time;
  asserts a pre-existing CREATE3 deployment's on-chain member set matches the configured one.
- **Tests** (+62): `test/prover/AggregatorProver.t.sol` (30), `test/prover/AggregatorProverIntegration.t.sol` (6,
  real `Portal` settlement), `test/scripts/AggregatorProverMemberValidation.t.sol` (26) — the first test
  coverage `scripts/Deploy.s.sol` has ever had.
- **Mocks:** `RevertingProver`, `MalformedProver`, `DirtyBitsProver`, `MockDomainProver`.
- **Docs:** `CLAUDE.md` gains the prover entry, `AGGREGATOR_PROVER_MEMBERS`, and
  `AGGREGATOR_PROVER_ALLOW_UNVERIFIED_MEMBERS`.

**Zero changes to any already-deployed contract.**

## Human Logic

```text
provenIntents(intentHash):
  FOR each member IN members, in immutable constructor order:
    IF member has no code             -> skip           # staticcall to codeless addr
                                                        # succeeds w/ empty returndata;
                                                        # decoding it reverts UNCATCHABLY
    staticcall member.provenIntents(intentHash)
    IF call reverted                  -> skip
    IF returndata length != 64        -> skip           # short/long payload; decoding
                                                        # reverts in OUR frame
    decode as (uint256, uint256)                        # cannot revert on 64 bytes
    IF high bits dirty                -> skip           # >>160 for address, >>64 for uint64
                                                        # also prevents silent truncation
                                                        # into a wrong-but-plausible claimant
    IF claimant == 0                  -> continue       # honest "no proof yet": fall through,
                                                        # do NOT terminate the search
    RETURN ProofData(claimant, destination)             # first non-zero claimant wins
  RETURN ProofData(0, 0)

INVARIANT: provenIntents never reverts for any member behaviour except an
           uncatchable OOG (oversized returndata, or an unbounded gas burn).
           This matters because IntentSource reads it from BOTH withdraw
           (:460) and _validateRefund (:880) — a revert freezes escrow for
           creator AND solver simultaneously.

challengeIntentProof(destination, routeHash, rewardHash):
  FOR each member IN members:
    IF member has no code -> skip
    try member.challengeIntentProof(...) catch -> skip
  emit ChallengeForwarded(hash)

INVARIANT: challengeIntentProof never reverts. IntentSource.withdraw calls it
           itself on the wrong-destination branch (:468), and batchWithdraw
           (:492) loops withdraw — so a revert here would abort payment for
           every unrelated intent in a batch.
```

## Logic Trace

```mermaid
flowchart TD
  A[IntentSource.withdraw] --> B[AggregatorProver.provenIntents]
  B --> C{member has code?}
  C -- no --> N[skip to next member]
  C -- yes --> D[staticcall provenIntents]
  D --> E{success?}
  E -- no --> N
  E -- yes --> F{returndata == 64 bytes?}
  F -- no --> N
  F -- yes --> G{high bits clean?}
  G -- no --> N
  G -- yes --> H{claimant != 0?}
  H -- no --> N
  H -- yes --> I[return first proof]
  N --> C
  I --> J{proof.destination == intent destination?}
  J -- yes --> K[pay claimant, status Withdrawn]
  J -- no --> L[forward challenge to ALL members, return without paying]
  L --> M[lying member deletes its own entry; honest entries untouched]
  M --> O[second withdraw returns honest proof and pays]
```

## Service Topology

```mermaid
flowchart LR
  subgraph src["Source chain"]
    Creator -->|"publish: reward.prover = AggregatorProver"| Portal_S[Portal / IntentSource]
    Portal_S -->|"provenIntents / challengeIntentProof"| Agg[AggregatorProver]
    Agg -->|"staticcall, priority order"| H[HyperProver]
    Agg -->|"staticcall"| M[MetaProver]
    Agg -->|"staticcall"| L[LayerZeroProver]
    Portal_S -->|"withdraw / refund"| Vault
  end
  subgraph dst["Destination chain"]
    Solver -->|"fulfillAndProve, picks a LIVE member"| Portal_D[Portal / Inbox]
    Portal_D -->|"prove"| HD[HyperProver_dst]
  end
  HD -.->|"bridge message"| H
  Agg -.->|"prove() reverts: dispatches nothing"| X((✗))
```

## Why This Shape

**Read-only, not dispatching.** `MessageBridgeProver.prove` is `only(PORTAL)` with `PORTAL`
immutable, so an aggregator cannot forward `prove()` to members without redeploying every member.
It also isn't needed: `Inbox.prove` is `public payable` and callable per-prover, so solvers already
choose their bridge. `AggregatorProver` is purely a source-chain read layer.

**Not extending `BaseProver`.** `BaseProver` supplies a `_provenIntents` mapping and
`_processIntentProofs` that would be permanently empty and unreachable here — dead state and dead
code in a contract that gates escrow release. `AggregatorProver` implements `IProver` directly and reuses
`Whitelist` for immutable member slots.

**Immutable membership, no admin.** `contracts/` contains zero `Ownable`/`AccessControl` contracts,
and this preserves that. Under 1-of-N the member set *is* the security parameter, so an owner able to
add a member would be an owner able to mint claimants.

**Stateless — no proof caching.** With immutable membership a cache would only save gas while adding
a hazard: a bogus proof latched from a compromised member would persist even after that member's own
entry was challenged away.

**Forwarding challenges rather than reverting.** `IntentSource.withdraw` calls `challengeIntentProof`
itself and then returns without paying. Reverting would make `withdraw` revert — and since
`batchWithdraw` loops `withdraw`, one planted bad proof would abort payment for every unrelated
intent in the batch. Forwarding confines the damage and self-heals in a second transaction.
Blanket-forwarding is precise despite the aggregator not knowing which member lied: each member
re-derives the hash and deletes **only its own** entry, and **only** on its own destination mismatch.

**No per-member gas cap.** A cap would silently skip an honest member whose read exceeded it,
leaving a solver who already delivered unpayable. Returndata *size* and *bit-range* checks are used
instead — neither can skip an honest member, since an honest member returns exactly 64 clean bytes.

**Trust model, stated plainly.** 1-of-N means the security floor is the **weakest** member. A
compromised member can forge a claimant and drain any vault whose intent names this prover. That is
inherent to the union semantics and is why membership is gated at deploy time and immutable
thereafter.

## Boundaries

**A known, unfixed defect ships with this PR — documented, not hidden.**

`AggregatorProver` introduces a bug class a single prover cannot have: **shadowing**. A single prover stores
exactly one `ProofData` per `intentHash`, so nothing can conceal anything. With priority-ordered
fan-out, a higher-priority member holding an entry whose `destination` is wrong conceals a valid
proof held by a lower-priority member. `withdraw` recovers (it forwards a challenge; a second
`withdraw` pays). `_validateRefund` does **not** — it reads the same shadowed value, never forwards a
challenge, and past `reward.deadline` refunds the **creator** while the solver who delivered goes
unpaid.

Pinned by a characterization test, `test_refund_shadowedProofRefundsCreator_knownLimitation`, which
asserts the current (wrong) behaviour so it cannot drift unnoticed.

Mitigated at deploy time, not fixed structurally. `validateAggregatorProverMembers` removes two of three
triggers: every member must have code, and must expose `chainIdByDomain(uint64)` — which only
`MessageBridgeProver` descendants do, and those are exactly the provers whose `destination` is
cross-checked against the bridge origin domain in `_handleCrossChainMessage`. `PolymerProver` writes
`_provenIntents` via its own `processIntent` and is rejected; `LocalProver` cannot satisfy the probe.

**Not included:**

- **The shadowing mechanism itself.** Two candidate fixes were rejected: agreement-preferred
  selection is useless in the expected two-member configuration, and a challenge-primed destination
  map is caller-supplied, letting an attacker prime a wrong destination and skip honest proofs —
  reproducing the wrongful refund it exists to prevent.
- **Domain-lane verification for escape-hatch members.** `AGGREGATOR_PROVER_ALLOW_UNVERIFIED_MEMBERS`
  skips it entirely. `CCIPProver` can only be admitted that way (separate deploy script) and uses a
  strict domain map with no fallback, so its domain table needs out-of-band verification.
- **The lane assertion is tautological for same-run members** — `deployHyperProver` builds the prover
  *from* `HYPER_DOMAIN_CONFIG` and the validator then compares against the same variable. It catches
  a prover **reused from an earlier run** with a different map, which is a real class of error, but
  not a wrong env var.
- **An oversized-returndata / unbounded-gas-burn member** can still force an uncatchable OOG.
  Documented in the contract.
- **A keeper** that detects member disagreement and calls the permissionless `challengeIntentProof`
  before `reward.deadline` — converting the shadowing loss into the two-transaction self-heal.
  Planned follow-up.
- **`refundTo` and ERC20 reward paths** are not separately covered; both funnel through
  `_refund` → `_validateRefund`, so the prover interaction is identical and token type never reaches
  the prover.

**Pre-deployment gate:** the exact ordered `AGGREGATOR_PROVER_MEMBERS` list per chain requires explicit
human sign-off. Order is priority and is consensus-critical; membership is immutable with no eject
path. `Deploy.s.sol` has never been exercised on a live network — a testnet dry-run is required.

## Validation

| Gate | Result |
| --- | --- |
| `forge test` | 702 passed / 0 failed / 0 skipped, 35 suites |
| `forge build` | clean, no solc warnings on new files |
| Per-task review | 6 tasks, each reviewed independently — 0 Critical, 0 Important |
| Whole-branch review | 0 Critical, 4 Important — all fixed in `038f3c7b` |
| Independent breakage review of the fix wave | no regression introduced |
| Freeze-fix review | ready to accept, 0 Critical, 0 Important |
| Escrow read path | verified revert-free for codeless, reverting, wrong-length, and dirty-bit members; empirically confirmed a two-`uint256` decode still reverts on a short payload, so the length guard is load-bearing |
| Rename verification | test functions 28/6/11 and assertions 39 identical pre/post; no functional line changed |
| Reentrancy | `provenIntents`: structurally impossible — `staticcall` places the whole subtree in a static context. `challengeIntentProof` is a plain `CALL`, so it is not covered by that argument; it is safe because the contract holds no state and no funds |

---

## Review round: changes since first review

Addressing @fish-sammy and @re1ro. One commit per issue, each independently
green so the range stays bisectable; suite is **720 passing**.

| Commit | Issue |
|---|---|
| `98659d90` | `_tryChainIdByDomain` strict decode (fish-sammy `:523`) |
| `3c98bb32` | `_tryProvenIntentsShape` empty-dynamic gap (re1ro §3) |
| `2200e61c` | unverified-members hatch into `DeploymentContext` (fish-sammy `:582`) |
| `e7cd8808` | explicit domain lanes for fallback resolvers (fish-sammy `:653`) |
| `2d57ab6e` | claimant decode via `AddressConverter` (fish-sammy `:250`) |
| `de5fd0fe` | reject codeless members at construction (fish-sammy `:311`) |
| `1c4986f9` | member set into the CREATE3 salt (re1ro §4/§5) |
| `7ea1ffd6` | fan-out gas + whitelist ordering pins (re1ro §6, nits) |

**Deploy-time probes** — both gate an immutable member set, so a miss is permanent:

- `_tryChainIdByDomain` decoded strictly to `uint64`. A member returning a word with
  non-zero upper bits made the probe itself revert in the script's own frame,
  uncatchably, aborting the deploy with a bare decode error instead of the intended
  require. It ran **unconditionally** on every member as the bridge-attestation gate,
  so this was on the mandatory path, not just the lane loop. Now decodes wide and
  range-checks.
- `_tryProvenIntentsShape` range-checked instead of requiring zeros, so a single empty
  dynamic return (`0x20`, `0x00`) passed as well-formed while `provenIntents` rejects
  that same payload at runtime — skipping the member for every intent hash, forever.

Both have regression pins verified to fail when the fix is reverted.

**Member-set integrity:**

- The ordered member set is mixed into the aggregator's CREATE3 salt. Previously a
  changed list silently reused the existing aggregator with its **old** members, and
  the only remedy was "change `SALT`" — which re-derives every prover address, so a
  list naming the canonical provers matched none of them. Validation was effectively
  first-deploy-only. Order is in the hash because order is priority.
- Members must have code on the target chain at construction (`MemberHasNoCode`); the
  runtime `code.length` guards are removed.
- `HyperProver`/`MetaProver` members require a non-empty domain config — their
  resolvers fall back to `chainId != 0 ? chainId : originDomain`, so an omitted lane
  records a wrong-but-plausible destination instead of failing closed.
  `LayerZeroProver` is exempt and that exemption is pinned.
- `AGGREGATOR_PROVER_ALLOW_UNVERIFIED_MEMBERS` moved into `DeploymentContext`, so the
  validator is a pure function of ctx and `vm.setEnv` is gone from the tests.

**Gas**, worst case at `MAX_MEMBERS` with no member holding the proof:
**31,799 cold / 15,762 warm**, pinned with tripwires.

### Still open, deliberately

- **Solver support is a hard prerequisite and does not exist yet.** `ProverType` has no
  `aggregator` and `ProverService.getProver()` resolves an aggregator address to `null`,
  so fulfilment fail-closes with `PROVER_UNSUPPORTED`. A `BaseProver` subclass is
  mandatory before an aggregator address is published anywhere the solver serves —
  adding the type to config alone would half-wire it, letting quotes advertise a prover
  the solver then refuses. Owned by @re1ro.
- **`sourceChainProver` is not validated on the send path.** Three of four solver lanes
  pass `reward.prover` straight into the bridge message recipient, so an intent naming
  the aggregator would address the proof to a contract with no `handle`/`lzReceive`.
  Being fixed solver-side; the contract-side guard is **not** in this PR because it
  would modify possibly-deployed provers.
- **Cross-chain member-set comparison.** `abi.encode(members)` is recorded per chain and
  never compared. The salt change removes the same-address-different-members case, but
  the address book still cannot show which set landed where.

### DIFF
--- a/contracts/prover/AggregatorProver.sol
+++ b/contracts/prover/AggregatorProver.sol
@@ -108,12 +108,44 @@ contract AggregatorProver is IProver, ERC165, Whitelist, Semver {
 
     /**
      * @notice Returns the first member proof with a non-zero claimant
+     * @dev Iterates members in immutable priority order. Members that are
+     *      codeless or revert are skipped, never propagated.
+     *
+     *      The `code.length` guard is load-bearing: a staticcall to a codeless
+     *      address SUCCEEDS with empty returndata, and ABI-decoding empty data
+     *      reverts in THIS frame where try/catch cannot catch it. Without it, a
+     *      member deployed only on other chains would brick withdraw AND refund
+     *      for every intent naming this aggregator. Mirrors the same defense at
+     *      IntentSource.sol:872-880.
+     *
+     *      No per-member gas cap by design: a cap would silently skip an honest
+     *      member whose read exceeds it, leaving a delivered solver unpayable
+     *      with no error. A compromised member can already forge a valid-looking
+     *      proof through the front door, so a returndata bomb grants it nothing
+     *      new. Fan-out is bounded by MAX_MEMBERS.
      * @param intentHash The intent hash to query
+     * @return First non-zero member proof, or a zero ProofData if none
      */
     function provenIntents(
         bytes32 intentHash
     ) external view returns (ProofData memory) {
-        intentHash; // silences unused-parameter warning until Task 2
+        bytes32[] memory members = getWhitelist();
+        uint256 length = members.length;
+
+        for (uint256 i = 0; i < length; ++i) {
+            address member = members[i].toAddress();
+
+            if (member.code.length == 0) continue;
+
+            try IProver(member).provenIntents(intentHash) returns (
+                ProofData memory proof
+            ) {
+                if (proof.claimant != address(0)) return proof;
+            } catch {
+                continue;
+            }
+        }
+
         return ProofData({claimant: address(0), destination: 0});
     }
 
--- a/contracts/prover/AggregatorProver.sol
+++ b/contracts/prover/AggregatorProver.sol
@@ -64,6 +64,11 @@ contract AggregatorProver is IProver, ERC165, Whitelist, Semver {
      *      Whitelist's constructor runs first, so a set larger than its own
      *      20-address cap reverts with WhitelistSizeExceeded, not
      *      InvalidMemberSet.
+     *
+     *      The `isValidAddress` check below is the ONLY reason the unguarded
+     *      `toAddress()` calls in getMembers/provenIntents/challengeIntentProof
+     *      never revert; loosening it would make all three revert-capable and
+     *      break IntentSource.batchWithdraw's per-intent isolation.
      * @param members Member prover addresses as bytes32, in priority order
      */
     constructor(bytes32[] memory members) Whitelist(members) {
--- a/contracts/prover/AggregatorProver.sol
+++ b/contracts/prover/AggregatorProver.sol
@@ -255,17 +255,21 @@ contract AggregatorProver is IProver, ERC165, Whitelist, Semver {
 
             if (!success || ret.length != 64) continue;
 
-            // Decode as two uint256 words first: unlike decoding directly to
+            // Decode to (bytes32, uint256) first: unlike decoding directly to
             // (address, uint64), this cannot revert for any 64-byte payload.
-            // Dirty high-order bits are then rejected by range check, exactly
-            // like a wrong-length payload, rather than left to solc's strict
-            // (address, uint64) decoder, which WOULD revert in this frame.
-            (uint256 rawClaimant, uint256 rawDestination) = abi.decode(
+            // Dirty high-order bits are then rejected by the checks below,
+            // exactly like a wrong-length payload, rather than left to solc's
+            // strict (address, uint64) decoder, which WOULD revert in this frame.
+            (bytes32 rawClaimant, uint256 rawDestination) = abi.decode(
                 ret,
-                (uint256, uint256)
+                (bytes32, uint256)
             );
 
-            if (rawClaimant >> 160 != 0 || rawDestination >> 64 != 0) {
+            // isValidAddress is the same >> 160 test AddressConverter applies
+            // everywhere else; using it keeps one canonical notion of "this
+            // word is an address" and makes the toAddress() below provably
+            // revert-free.
+            if (!rawClaimant.isValidAddress() || rawDestination >> 64 != 0) {
                 continue;
             }
 
@@ -278,7 +282,7 @@ contract AggregatorProver is IProver, ERC165, Whitelist, Semver {
             // _resolveChainId reverts UnregisteredDomain(0).
             if (rawDestination == 0) continue;
 
-            address claimant = address(uint160(rawClaimant));
+            address claimant = rawClaimant.toAddress();
             uint64 destination = uint64(rawDestination);
 
             if (claimant != address(0)) {
--- a/contracts/prover/AggregatorProver.sol
+++ b/contracts/prover/AggregatorProver.sol
@@ -24,6 +24,12 @@ import {AddressConverter} from "../libs/AddressConverter.sol";
  *      repo Whitelist holds REMOTE cross-VM prover addresses; here it holds
  *      LOCAL EVM member contracts.
  */
+/// @dev Whitelist's getWhitelist()/getWhitelistSize()/isWhitelisted() carry a
+///      DIFFERENT meaning here than on every other IProver implementation: on
+///      those, "whitelisted" means a trusted remote cross-VM prover address;
+///      here it means a local EVM member of the 1-of-N union. Off-chain
+///      tooling that reads isWhitelisted() uniformly across IProver
+///      implementations will misread this contract.
 contract AggregatorProver is IProver, ERC165, Whitelist, Semver {
     using AddressConverter for bytes32;
 
@@ -114,20 +120,57 @@ contract AggregatorProver is IProver, ERC165, Whitelist, Semver {
     /**
      * @notice Returns the first member proof with a non-zero claimant
      * @dev Iterates members in immutable priority order. Members that are
-     *      codeless or revert are skipped, never propagated.
+     *      codeless, revert, or return a zero claimant are skipped and never
+     *      propagated: a zero-claimant success must fall through to the next
+     *      member, not terminate the search.
      *
-     *      The `code.length` guard is load-bearing: a staticcall to a codeless
-     *      address SUCCEEDS with empty returndata, and ABI-decoding empty data
-     *      reverts in THIS frame where try/catch cannot catch it. Without it, a
-     *      member deployed only on other chains would brick withdraw AND refund
-     *      for every intent naming this aggregator. Mirrors the same defense at
-     *      IntentSource.sol:872-880.
+     *      The `code.length` guard closes the CODELESS case: a staticcall to a
+     *      codeless address SUCCEEDS with empty returndata, and ABI-decoding
+     *      empty data would revert in THIS frame where try/catch cannot catch
+     *      it. Without it, a member deployed only on other chains would brick
+     *      withdraw AND refund for every intent naming this aggregator.
+     *      Mirrors the same defense at IntentSource.sol:872-880.
+     *
+     *      The call below is a low-level staticcall, not an interface call,
+     *      and checks the exact returndata size before decoding. That closes
+     *      the WRONG-SHAPE case: a code-bearing member that returns SUCCESS
+     *      with insufficient returndata would otherwise make solc's generated
+     *      decoder revert in THIS frame, outside any try/catch — not
+     *      griefing, but a permanent freeze of both withdraw and refund for
+     *      every intent naming this aggregator, since provenIntents is read by
+     *      both. A non-dynamic ProofData{address; uint64;} ABI-encodes to
+     *      exactly 64 bytes, so `ret.length == 64` is the exact-shape check.
+     *
+     *      An OVERSIZED returndata bomb is still possible and is ACCEPTED:
+     *      solc copies the full returndata into memory in our frame before
+     *      any length check runs, so a member returning gigabytes of data
+     *      still costs us quadratic memory-expansion gas, an uncatchable OOG.
+     *      This is tolerated because deploy-time validation
+     *      (`Deploy.validateAggregatorMembers`) restricts members to
+     *      `MessageBridgeProver` descendants built from this repo — not
+     *      arbitrary bytecode. That validator also assumes NON-PROXY member
+     *      bytecode: a proxy whose fallback returned a plausible 64-byte
+     *      payload could defeat the shape checks above; it does not defend
+     *      against a malicious proxy member.
      *
      *      No per-member gas cap by design: a cap would silently skip an honest
      *      member whose read exceeds it, leaving a delivered solver unpayable
      *      with no error. A compromised member can already forge a valid-looking
      *      proof through the front door, so a returndata bomb grants it nothing
      *      new. Fan-out is bounded by MAX_MEMBERS.
+     *
+     *      KNOWN LIMITATION (shadowing): a member holding an entry whose
+     *      `destination` is wrong shadows a valid proof held by a
+     *      lower-priority member, since this function returns the first
+     *      non-zero claimant. `IntentSource.withdraw` recovers by forwarding a
+     *      challenge on its wrong-destination branch, so a second `withdraw`
+     *      pays, but `_validateRefund` reads the same shadowed value, never
+     *      forwards a challenge, and past `reward.deadline` refunds the
+     *      creator while the solver who delivered goes unpaid. The mitigation
+     *      is `Deploy.validateAggregatorMembers`, which restricts members to
+     *      provers whose `destination` is bridge-attested by
+     *      `MessageBridgeProver._handleCrossChainMessage`; the asymmetry
+     *      itself remains in `IntentSource`.
      * @param intentHash The intent hash to query
      * @return First non-zero member proof, or a zero ProofData if none
      */
@@ -142,12 +185,23 @@ contract AggregatorProver is IProver, ERC165, Whitelist, Semver {
 
             if (member.code.length == 0) continue;
 
-            try IProver(member).provenIntents(intentHash) returns (
-                ProofData memory proof
-            ) {
-                if (proof.claimant != address(0)) return proof;
-            } catch {
-                continue;
+            (bool success, bytes memory ret) = member.staticcall(
+                abi.encodeWithSelector(
+                    IProver.provenIntents.selector,
+                    intentHash
+                )
+            );
+
+            if (!success || ret.length != 64) continue;
+
+            (address claimant, uint64 destination) = abi.decode(
+                ret,
+                (address, uint64)
+            );
+
+            if (claimant != address(0)) {
+                return
+                    ProofData({claimant: claimant, destination: destination});
             }
         }
 
--- a/contracts/prover/AggregatorProver.sol
+++ b/contracts/prover/AggregatorProver.sol
@@ -63,6 +63,9 @@ contract AggregatorProver is IProver, ERC165, Whitelist, Semver {
     /// @notice Member appears more than once in the set
     error DuplicateMember(bytes32 member);
 
+    /// @notice Thrown when a member has no code on this chain at construction
+    error MemberHasNoCode(address member);
+
     /// @notice Emitted once per member at construction, so the immutable,
     ///         setter-less set is auditable after deploy
     /// @param member Member prover address
@@ -109,7 +112,29 @@ contract AggregatorProver is IProver, ERC165, Whitelist, Semver {
                 }
             }
 
-            emit MemberRegistered(member.toAddress(), i);
+            address memberAddr = member.toAddress();
+
+            // Fail closed here so the read paths do not have to guard at
+            // runtime. A codeless member would otherwise be skipped forever by
+            // provenIntents (silently shrinking the trust set to fewer members
+            // than the operator believes they configured) and would revert
+            // challengeIntentProof outright, since solc's extcodesize check on
+            // a void high-level call lands in THIS frame, outside the catch.
+            //
+            // Checking once at construction is sound because code cannot
+            // disappear afterwards: post-EIP-6780 SELFDESTRUCT only clears code
+            // when it runs in the same transaction that created the contract,
+            // and no chain this aggregator targets predates Cancun. There is no
+            // AggregatorProverTron variant; TRON deploys via scripts/tron.
+            //
+            // Consequence, deliberately accepted: on a chain where a member is
+            // not live yet, deployment REVERTS rather than silently deploying a
+            // smaller set at the same address.
+            if (memberAddr.code.length == 0) {
+                revert MemberHasNoCode(memberAddr);
+            }
+
+            emit MemberRegistered(memberAddr, i);
         }
     }
 
@@ -244,8 +269,10 @@ contract AggregatorProver is IProver, ERC165, Whitelist, Semver {
         for (uint256 i = 0; i < length; ++i) {
             address member = members[i].toAddress();
 
-            if (member.code.length == 0) continue;
-
+            // No code.length guard: the constructor already rejects codeless
+            // members, and a staticcall to a codeless address returns success
+            // with EMPTY returndata anyway, which the ret.length != 64 check
+            // below rejects. Dropping it saves an EXTCODESIZE per member.
             (bool success, bytes memory ret) = member.staticcall(
                 abi.encodeWithSelector(
                     IProver.provenIntents.selector,
@@ -321,15 +348,15 @@ contract AggregatorProver is IProver, ERC165, Whitelist, Semver {
         for (uint256 i = 0; i < length; ++i) {
             address member = members[i].toAddress();
 
-            // Load-bearing, not defensive filler: solc's extcodesize check
-            // inside a high-level `try` call reverts in THIS frame when the
-            // target is codeless, outside any `catch` — the `try/catch`
-            // below only guards the CALL itself, not the codeless-target
-            // case. Without this, a member deployed only on other chains
-            // would revert this whole function, and the wrong-destination
-            // branch in IntentSource.withdraw that calls it would revert too.
-            if (member.code.length == 0) continue;
-
+            // No code.length guard: the constructor rejects codeless members and
+            // code cannot disappear afterwards (EIP-6780), so the codeless case
+            // is unreachable. It is worth stating what it would cost if that
+            // assumption ever broke: solc's extcodesize check on this void
+            // high-level call reverts in THIS frame, outside the `catch` below,
+            // so this whole function would revert — and with it the
+            // wrong-destination branch of IntentSource.withdraw that calls it,
+            // breaking batchWithdraw's per-intent isolation. The `try/catch`
+            // guards the CALL, never the codeless-target case.
             try
                 IProver(member).challengeIntentProof(
                     destination,
--- a/contracts/prover/AggregatorProver.sol
+++ b/contracts/prover/AggregatorProver.sol
@@ -132,14 +132,23 @@ contract AggregatorProver is IProver, ERC165, Whitelist, Semver {
      *      Mirrors the same defense at IntentSource.sol:872-880.
      *
      *      The call below is a low-level staticcall, not an interface call,
-     *      and checks the exact returndata size before decoding. That closes
-     *      the WRONG-SHAPE case: a code-bearing member that returns SUCCESS
-     *      with insufficient returndata would otherwise make solc's generated
-     *      decoder revert in THIS frame, outside any try/catch — not
-     *      griefing, but a permanent freeze of both withdraw and refund for
-     *      every intent naming this aggregator, since provenIntents is read by
-     *      both. A non-dynamic ProofData{address; uint64;} ABI-encodes to
-     *      exactly 64 bytes, so `ret.length == 64` is the exact-shape check.
+     *      and the read path is revert-free for ANY 64-byte payload, honest or
+     *      not. A non-dynamic ProofData{address; uint64;} ABI-encodes to
+     *      exactly 64 bytes, so `ret.length == 64` closes the WRONG-SHAPE case
+     *      first: a code-bearing member returning SUCCESS with insufficient
+     *      returndata would otherwise make solc's generated decoder revert in
+     *      THIS frame, outside any try/catch. That length check alone is not
+     *      enough, though: decoding a full 64-byte payload directly to
+     *      (address, uint64) is ALSO strict — solc reverts, still in this
+     *      frame, if the upper 96 bits of the address word or the upper 192
+     *      bits of the uint64 word are non-zero. So the 64 bytes are decoded
+     *      first as (uint256, uint256), which cannot revert for any bit
+     *      pattern, and only then range-checked (`>> 160`/`>> 64` non-zero)
+     *      before narrowing; a dirty payload is treated exactly like a
+     *      wrong-length one, i.e. skipped via `continue`. Either failure
+     *      mode — wrong length or dirty high bits — would otherwise be a
+     *      permanent freeze of both withdraw and refund for every intent
+     *      naming this aggregator, since provenIntents is read by both.
      *
      *      An OVERSIZED returndata bomb is still possible and is ACCEPTED:
      *      solc copies the full returndata into memory in our frame before
@@ -194,10 +203,22 @@ contract AggregatorProver is IProver, ERC165, Whitelist, Semver {
 
             if (!success || ret.length != 64) continue;
 
-            (address claimant, uint64 destination) = abi.decode(
+            // Decode as two uint256 words first: unlike decoding directly to
+            // (address, uint64), this cannot revert for any 64-byte payload.
+            // Dirty high-order bits are then rejected by range check, exactly
+            // like a wrong-length payload, rather than left to solc's strict
+            // (address, uint64) decoder, which WOULD revert in this frame.
+            (uint256 rawClaimant, uint256 rawDestination) = abi.decode(
                 ret,
-                (address, uint64)
+                (uint256, uint256)
             );
+
+            if (rawClaimant >> 160 != 0 || rawDestination >> 64 != 0) {
+                continue;
+            }
+
+            address claimant = address(uint160(rawClaimant));
+            uint64 destination = uint64(rawDestination);
 
             if (claimant != address(0)) {
                 return
--- a/contracts/prover/AggregatorProver.sol
+++ b/contracts/prover/AggregatorProver.sol
@@ -8,7 +8,7 @@ import {Whitelist} from "../libs/Whitelist.sol";
 import {AddressConverter} from "../libs/AddressConverter.sol";
 
 /**
- * @title EcoProver
+ * @title AggregatorProver
  * @notice Reports an intent as proven when ANY member prover has proven it
  * @dev Stateless 1-of-N union over an immutable set of member provers. Records
  *      no proofs of its own and dispatches no messages: solvers prove through a
@@ -35,13 +35,13 @@ import {AddressConverter} from "../libs/AddressConverter.sol";
 ///      recipient. `data.sourceChainProver` in a `Portal.prove` /
 ///      `fulfillAndProve` call must be one of `getMembers()` resolved on the
 ///      SOURCE chain, never `reward.prover` (which may legitimately be this
-///      aggregator). `EcoProver` implements no `handle`, no `lzReceive`, and
+///      aggregator). `AggregatorProver` implements no `handle`, no `lzReceive`, and
 ///      no fallback, so delivery to it reverts on an unknown selector
 ///      forever — permanently stranding the message. For every other prover,
 ///      CREATE3 parity makes `reward.prover` the correct recipient too, which
 ///      makes this a real footgun: the pattern that works for every other
 ///      prover silently does not work here.
-contract EcoProver is IProver, ERC165, Whitelist, Semver {
+contract AggregatorProver is IProver, ERC165, Whitelist, Semver {
     using AddressConverter for bytes32;
 
     /// @notice Proof mechanism identifier
@@ -168,7 +168,7 @@ contract EcoProver is IProver, ERC165, Whitelist, Semver {
      *      still costs us quadratic memory-expansion gas; a member that
      *      simply burns gas without returning is the same failure mode by a
      *      different mechanism. This is tolerated because deploy-time
-     *      validation (`Deploy.validateEcoProverMembers`) only probes that
+     *      validation (`Deploy.validateAggregatorProverMembers`) only probes that
      *      each member exposes `chainIdByDomain(uint64)` — a duck-typed check
      *      that any contract implementing that one function passes, not a
      *      guarantee of `MessageBridgeProver`-descended or repo-built
@@ -212,7 +212,7 @@ contract EcoProver is IProver, ERC165, Whitelist, Semver {
      *      pays, but `_validateRefund` reads the same shadowed value, never
      *      forwards a challenge, and past `reward.deadline` refunds the
      *      creator while the solver who delivered goes unpaid. The mitigation
-     *      is `Deploy.validateEcoProverMembers`, which restricts members to
+     *      is `Deploy.validateAggregatorProverMembers`, which restricts members to
      *      provers whose `destination` is bridge-attested by
      *      `MessageBridgeProver._handleCrossChainMessage`; the asymmetry
      *      itself remains in `IntentSource`.
--- a/contracts/prover/AggregatorProver.sol
+++ b/contracts/prover/AggregatorProver.sol
@@ -150,13 +150,45 @@ contract AggregatorProver is IProver, ERC165, Whitelist, Semver {
     }
 
     /**
-     * @notice Forwards a challenge to every member
+     * @notice Forwards a challenge to every member prover
+     * @dev Forwarding is required, not cosmetic: IntentSource.withdraw calls
+     *      this itself on its wrong-destination branch (IntentSource.sol:468).
+     *      Reverting here would revert withdraw, and would revert an entire
+     *      batchWithdraw (IntentSource.sol:492) over one planted bad proof.
+     *
+     *      Blanket-forwarding is precise even though this contract cannot tell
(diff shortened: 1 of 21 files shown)


re1ro (commented):
Reviewing from the solver side — @fish-sammy owns the Solidity correctness pass, so I've concentrated on whether eco-solver can actually consume this and on the deploy path. The design reads well: stateless, no owner, no setter, immutable membership, and the NatSpec at `AggregatorProver.sol:34-43` already names the sharpest hazard rather than leaving it implicit. Most of what follows is **our** problem, not yours.

---

## 1. The `sourceChainProver` footgun your NatSpec warns about is live today, not theoretical

This is the one I'd want closed before an aggregator address is published anywhere.

`prove()` reverting `ProvingNotSupported` catches a misrouted `prover` argument, but it does not catch a wrong `sourceChainProver` buried in the `data` blob — and that is the field that becomes the bridge message recipient. Nothing validates it on the send path:

- `HyperProver.sol:238` — `params.recipientAddress = unpacked.sourceChainProver;`
- `MetaProver.sol:264` — `recipient = sourceChainProver;`
- `LayerZeroProver.sol:294` — `params.receiver = unpacked.sourceChainProver;`

**Why it's not theoretical: eco-solver already passes `reward.prover` straight into that field.** Three of our four bridge lanes do it —

- `hyper.prover.ts:31` — `[[intent.reward.prover as Hex, '0x', zeroAddress]]`
- `metalayer.prover.ts:35` — `AddressNormalizer.toEvmBytes32(intent.reward.prover)`
- `layerzero.prover.ts:49` — same

Only `ccip.prover.ts:28` resolves per-type via `getProverAddress(sourceChainId, this.type)`.

So the first intent naming the AggregatorProver as `reward.prover` has three of our four lanes address the proof message *to the aggregator* — which implements no `handle` / `lzReceive` / fallback by design. The message is undeliverable forever, the goods are already delivered on the destination chain, and the fulfillment is a total loss with no recovery path. Right now the NatSpec warning is the only thing standing between us and that.

**Two asks, and I'd take either:**

1. **I'll own the solver-side fix** — resolve `sourceChainProver` per prover type (the CCIP pattern) instead of from `reward.prover`. That's ours regardless of what you do here.
2. **A belt-and-braces guard on the send path.** All three provers already hold a `Whitelist` and enforce it on the *receive* side (`MessageBridgeProver.sol:150`), and all three inherit `validateWhitelisted` (`Whitelist.sol:172`) — so it's roughly one line each. Out of this PR's diff, so entirely reasonable to split out, but this PR is what makes it reachable.

---

## 2. eco-solver cannot consume an aggregator-named intent at all today

Stating this so it's tracked, not as a defect here — the solver's behaviour is correctly fail-closed.

`ProverType` (`prover.interface.ts:7-15`) has no `aggregator`, and `ProverService.getProver()` (`prover.service.ts:112-124`) matches an address by iterating *registered instances*, so an aggregator address resolves to `null`. Fulfillment then rejects cleanly and typed: `ProverSupportValidation` → `PROVER_UNSUPPORTED` → PERMANENT / REPORT_CONFIG / CONFIGURATION. Quotes don't fail — they just always come back naming a concrete prover, never the aggregator.

So the PR's stated rationale ("creator names a set of bridges, solver picks a live one") is blocked on solver work that doesn't exist yet. Your **Pre-deployment gate** covers the member list; it doesn't mention solver support. Worth adding.

**One trap for whoever picks that ticket up** — config alone is not enough, and it half-wires in a way that's worse than a plain reject:

- `EvmConfigService.getAvailableProvers` is pure `Object.entries(network.provers)`, so merely adding `aggregator` to `ProverTypeValues` lets `selectProverForRoute` **select** it and `resolvePortalAndProver` **advertise** it in a quote — while `getProver()` still returns `null`, because it keys off registered instances. We'd quote intents naming a prover we then refuse to fulfill with a PERMANENT error.
- `ChainListener` would start realtime + durable `IntentProven` pollers on the aggregator address (`chain.listener.ts:503-505`, `:687-688` iterate `provers` with no type filter) — on a contract that never emits it. Confirmed against the PR: `AggregatorProver` declares only `MemberRegistered` (:70) and `ChallengeForwarded` (:75).
- `assertProverAddressUniqueness` also iterates registered instances, so the boot-time collision guard silently wouldn't cover it.

A new `BaseProver` subclass is mandatory, not optional. I'll write that up on our side.

---

## 3. `_tryProvenIntentsShape` waves through the one payload class the runtime guard exists for

`Deploy.s.sol:557` returns `rawClaimant >> 160 == 0 && rawDestination >> 64 == 0`.

A member whose `provenIntents(bytes32)` returns a single empty dynamic value (`bytes`/`string`/any array) ABI-encodes to exactly 64 bytes — offset head `0x20`, then length `0x00`. So `rawClaimant == 0x20` (passes `>>160`) and `rawDestination == 0` (passes `>>64`), and the probe calls it well-formed. At runtime `AggregatorProver.provenIntents` rejects that identical payload via its `rawDestination == 0` guard (`AggregatorProver.sol:263`) — so the member is skipped for **every** intentHash, forever, and membership is immutable.

That's against the probe's own `@dev` at `:531-535` ("membership is immutable, so this is the last point such a member can be caught"). Of the two 64-byte malformed classes the runtime guard handles, the probe catches dirty-bits and misses empty-dynamic.

Nothing in this repo triggers it — every prover here returns the static `ProofData` — so this is hardening against a future or third-party member, realistically one added via `ALLOW_UNVERIFIED_MEMBERS`. Worth fixing anyway because the deployment is immutable and the fix is one line:

```solidity
return rawClaimant == 0 && rawDestination == 0;
```

The probe uses `bytes32(0)`, an unproven hash, so an honest member returns exactly a zero `ProofData`. I checked out `56f8ea2`, applied that line and re-ran: **18/18** in `AggregatorProverMemberValidation.t.sol` and **53/53** across the three aggregator suites still pass, and the empty-dynamic payload now fails the probe (`0x20 >> 160 == 0` passes today; `0x20 == 0` does not).

---

## 4. Member-set validation is disabled by construction after the first aggregator deploy

`Deploy.s.sol:703`/`:708` tell the operator to "change SALT" when the on-chain member set disagrees with `AGGREGATOR_PROVER_MEMBERS`. But `SALT` is the single root salt for the whole stack (`ctx.salt` at `:83`, every prover salt via `getContractSalt(ctx.salt, ...)` at `:178-201`), and there's no per-contract salt override anywhere in `scripts/`.

So changing `SALT` re-derives `ctx.hyperProver` / `ctx.metaProver` / `ctx.layerZeroProver`, a member list naming the canonical in-use provers matches none of them, and the deploy hard-fails unless `AGGREGATOR_PROVER_ALLOW_UNVERIFIED_MEMBERS=true`. Net effect: the validator only ever works on the *first* aggregator deploy, and is off by construction for exactly the deploys that mutate a consensus-critical immutable set.

To be precise about what the hatch actually skips, since the docs are a bit loose here: the unconditional guards still run — EVM shape (:589), non-zero (:592), has-code (:597), the `chainIdByDomain` presence probe (:623-627, which *is* the mitigation the shadowing NatSpec names), and the `provenIntents` shape probe. What's skipped is only the per-lane `resolved == domains[j].chainId` loop (:653-666), which is frequently zero-iteration for Hyper/Meta. The real loss is on **LayerZero** members, whose strict domain→chainId map has no fallback.

Mixing `keccak256(abi.encode(members))` into the aggregator's CREATE3 salt would fix this properly — a member-set change then re-derives the aggregator address on its own, without touching the root salt or its members. No test covers the salt-change path today; `AggregatorProverMemberValidation.t.sol` only toggles the env var directly (:139, :153, :156).

---

## 5. Per-chain member sets are recorded and then discarded

The aggregator's CREATE3 address ignores bytecode and constructor args, and `SALT` is one version-derived root salt reused on every chain — so the aggregator lands at the **same address on every chain regardless of its member set**, while the member set is a constructor arg that can legitimately differ per chain. The readback at `:698-711` is same-chain only, and `validateAggregatorProverMembers` validates each member's properties, never the list's completeness or its **order** — and order is priority (`AggregatorProver.sol:228-274` returns the first non-zero claimant).

The data to catch this already exists and is thrown away: `:312-318` records `abi.encode(members)` next to `block.chainid`, `writeDeployFile` writes it, and `sr-deploy-contracts.ts:319-342` concatenates every chain's rows into `out/deployedAll.csv` — never compared. `processContractsForJson` then emits addresses only, so the published address book genuinely can't tell `[Hyper, Meta]` from `[Hyper]` from `[Meta, Hyper]`.

Realistic divergence vectors are the manual per-chain `forge script` the README documents, and a chain onboarded mid-release where a member isn't live yet — `require(member.code.length > 0)` forces the operator to drop it for that chain, producing a legitimately smaller set at an identical address.

---

## 6. Ours to fix, listing so it's visible

- **We drop `ProofData.destination` on both proof-read paths** (`evm.reader.service.ts:770-772`, and the TVM twin), so the shadowing case this PR documents is indistinguishable from a normal proof on our side. Ours.
- **A phantom `provenEvent` coupled to a real payout block** suppresses our only paging alert for that window (`intent.repository.ts:955`). Ours, and I'll ticket it.
- **Gas**: measured **+49k per withdraw** at `MAX_MEMBERS` for the `provenIntents` fan-out. Not a problem, but our blind-fallback withdrawal gas floor is a global 150k, so I want the number written down somewhere before we raise members.

## Nits

- `AggregatorProver.sol:313` — the validation row "Reentrancy: structurally impossible" should be scoped to `provenIntents`; `challengeIntentProof` is a plain `CALL`.
- `Deploy.s.sol:698` — the already-deployed member-set guard has no test.
- The PR description documents an earlier revision of the branch (pre-`7ad38d7`), including a safety guard that has since changed.

---

**Net:** I'm not blocking — the contract itself is clean and the fail-closed behaviours are right. The two I'd genuinely like resolved before an aggregator address goes live anywhere the solver serves are **§1** (I'll own the solver half regardless) and **§3** (one line). §4 and §5 are deploy-ergonomics on an immutable, consensus-critical set, so worth doing while it's cheap.

Also checked and found nothing wrong with: the differing-claimants-across-members question (priority order is documented and deterministic), `challengeIntentProof`'s try/catch swallowing, the 1-of-N weakest-member trade (documented), return-data-size handling from a malicious member, and duplicate/zero/`MAX_MEMBERS` boundary validation — the `DirtyBitsProver` / `MalformedProver` / `RevertingProver` / `EmptyDynamicProver` test doubles cover more than I expected.

Happy to pair on the solver-side prover subclass if that unblocks your rollout.

contracts/prover/AggregatorProver.sol
```diff
@@ -0,0 +1,361 @@
...
+
+            // Decode as two uint256 words first: unlike decoding directly to
+            // (address, uint64), this cannot revert for any 64-byte payload.
+            // Dirty high-order bits are then rejected by range check, exactly
+            // like a wrong-length payload, rather than left to solc's strict
+            // (address, uint64) decoder, which WOULD revert in this frame.
+            (uint256 rawClaimant, uint256 rawDestination) = abi.decode(
+                ret,
+                (uint256, uint256)
+            );
```
fish-sammy, line 293:
```suggestion
            (bytes32 rawClaimant, uint256 rawDestination) = abi.decode(ret, (bytes32, uint256));
```
so that you can use `AddressConverter` to check `rawClaimant` validity and perform the conversion.

contracts/prover/AggregatorProver.sol
```diff
@@ -0,0 +1,361 @@
...
+     * @return First non-zero member proof, or a zero ProofData if none
+     */
+    function provenIntents(
+        bytes32 intentHash
+    ) external view returns (ProofData memory) {
+        bytes32[] memory members = getWhitelist();
+        uint256 length = members.length;
+
+        for (uint256 i = 0; i < length; ++i) {
+            address member = members[i].toAddress();
```
fish-sammy, line 270: Since members have to be address, should just pass them in as address[] instead of bytes32[] in the constructor.

contracts/prover/AggregatorProver.sol
```diff
@@ -0,0 +1,361 @@
...
+
+            // Load-bearing, not defensive filler: solc's extcodesize check
+            // inside a high-level `try` call reverts in THIS frame when the
+            // target is codeless, outside any `catch` — the `try/catch`
+            // below only guards the CALL itself, not the codeless-target
+            // case. Without this, a member deployed only on other chains
+            // would revert this whole function, and the wrong-destination
+            // branch in IntentSource.withdraw that calls it would revert too.
+            if (member.code.length == 0) continue;
```
fish-sammy: This should be checked in the constructor.

scripts/Deploy.s.sol
```diff
@@ -441,6 +493,233 @@ contract Deploy is Script {
...
+        address member,
+        uint64 domain
+    ) internal view returns (bool ok, uint64 chainId) {
+        (bool success, bytes memory ret) = member.staticcall(
+            abi.encodeWithSignature("chainIdByDomain(uint64)", domain)
+        );
+        if (!success || ret.length != 32) {
+            return (false, 0);
+        }
+        return (true, abi.decode(ret, (uint64)));
```
fish-sammy: `abi.decode(ret, (uint64))` is strict — solc reverts if the upper 192 bits of the returned word are non-zero, and that revert lands in **this** frame. That is exactly the trap the `@dev` at :502-507 says checking `success` and `ret.length == 32` "avoids entirely". It doesn't: the length check only covers the *short/long* payload class, not the dirty-bits class.

Concrete failure: a member whose `chainIdByDomain` returns a `uint256` (or anything with dirty high bits) aborts the entire deploy with a bare decode revert, instead of returning `(false, 0)` and surfacing the intended `"member does not expose chainIdByDomain; destination not bridge-attested"` require. Verified on forge 1.5.1 / solc 0.8.35 — a callee returning `abi.encode(type(uint256).max)` makes the probing function itself revert, uncatchably from the caller.

It matters here because `_tryChainIdByDomain` is the *unconditional* gate applied to every member, including third-party ones admitted via `AGGREGATOR_PROVER_ALLOW_UNVERIFIED_MEMBERS` — the only members whose return shape you don't control.

`_tryProvenIntentsShape` 30 lines below already does this correctly. Same pattern here:

```solidity
uint256 raw = abi.decode(ret, (uint256));
if (raw >> 64 != 0) {
    return (false, 0);
}
return (true, uint64(raw));
```

scripts/Deploy.s.sol
```diff
@@ -441,6 +493,233 @@ contract Deploy is Script {
...
+                domains = ctx.layerZeroDomainConfig;
+            } else {
+                require(
+                    allowUnverified,
+                    "member is not a prover deployed in this run"
+                );
+                continue;
+            }
+
+            for (uint256 j = 0; j < domains.length; j++) {
```
fish-sammy, line 708: This loop is the only thing that verifies a member's domain→chainId map, and it runs **zero times** for `HyperProver`/`MetaProver` whenever `HYPER_DOMAIN_CONFIG`/`META_DOMAIN_CONFIG` is empty — which `scripts/README.md` and `CLAUDE.md` both call the normal case ("exceptions-only … may legitimately be empty").

Pair that with `HyperProver._resolveChainId` (`HyperProver.sol:106-108`): `chainId != 0 ? chainId : originDomain`. An unregistered Hyperlane domain `D` doesn't fail closed — it records `destination = D`, a non-zero, plausible-looking wrong chain id. `MetaProver` has the same fallback; only `LayerZeroProver` inherits the strict `MessageBridgeProver` base that reverts `UnregisteredDomain`.

So the `@dev` at :566-576 (and the matching NatSpec on `AggregatorProver.provenIntents`) overstates the mitigation. For the two most likely members, "destination is bridge-attested" reduces to "the bridge's domain id happens to equal the chain id".

Scenario, no compromise required: members `[HyperProver, MetaProver]`, destination chain whose Hyperlane domain != its chainId and which was omitted from `HYPER_DOMAIN_CONFIG`. Solver fulfils and proves over both lanes. `HyperProver` stores `destination = D` (wrong) and, being index 0, shadows `MetaProver`'s correct entry. After `reward.deadline`, `_validateRefund` reads the shadowed value, forwards no challenge, and refunds the creator while the solver who delivered goes unpaid — the exact `test_refund_shadowedProofRefundsCreator_knownLimitation` path, reached by a config omission rather than a compromise.

Two ways out, either is fine: require a non-empty domain config for any member matching `ctx.hyperProver`/`ctx.metaProver`, or soften the claim to "removes the codeless case; the domain-map case is only covered for members with an explicit lane config".

scripts/Deploy.s.sol
```diff
@@ -441,6 +493,233 @@ contract Deploy is Script {
...
+     *      bridge-attested by MessageBridgeProver._handleCrossChainMessage,
+     *      removing the two config-triggered causes (codeless member,
+     *      non-bridge-attested member). The asymmetry itself remains in
+     *      IntentSource.
+     * @param ctx Deployment context carrying the member list and domain configs
+     */
+    function validateAggregatorProverMembers(
+        DeploymentContext memory ctx
+    ) internal view {
+        bool allowUnverified = vm.envOr(
```
fish-sammy: Every other deploy input is read once in `run()` into `DeploymentContext`; this one is read from the environment deep inside a validator. That is what forces the tests to drive it with `vm.setEnv` (`AggregatorProverMemberValidation.t.sol:139/153/156`) — process-wide mutable state — and the comment above that test already documents an observed cross-test race it worked around by merging two cases into one function.

The workaround leaves a sharper hazard than the race it fixed: if `harness.exposedValidate(ctx)` at :154 reverts, :156 never runs and `AGGREGATOR_PROVER_ALLOW_UNVERIFIED_MEMBERS` stays `"true"` for the remainder of the process. Every later test that depends on the hatch being *off* — `test_rejectsDomainMapMismatch`, and the negative half of this one on a re-run — then passes for the wrong reason. One genuine failure turns into a cascade of false greens on an immutable, consensus-critical member set.

Reading it in `run()` into a `ctx.allowUnverifiedMembers` field fixes both: `validateAggregatorProverMembers` becomes a pure function of `ctx`, and the tests set a struct field instead of the environment.

fish-sammy (commented):
Contract read path looks solid — the revert-free fan-out, the codeless/length/dirty-bits/zero-destination ladder, and the challenge forwarding all hold up under the malicious-member cases I could construct. Comments inline; three of them are on `Deploy.s.sol`, which is where the remaining risk is concentrated (immutable member set, so a deploy-time miss is permanent).

The one I'd want fixed before this ships:

- **`_tryChainIdByDomain` (:523)** — `abi.decode(ret, (uint64))` is strict and reverts in the script's own frame on dirty high bits, which is precisely the trap its own `@dev` claims the `ret.length == 32` check avoids. Verified on forge 1.5.1 / solc 0.8.35. `_tryProvenIntentsShape` right below already does it correctly; copy that.

The other two:

- **Domain-lane loop (:653)** — it runs zero times for `HyperProver`/`MetaProver` with an empty domain config (the documented normal case), and `_resolveChainId`'s `domain == chainId` fallback then records a wrong-but-plausible `destination` rather than failing closed. That makes the shadowing→wrongful-refund path reachable from a config omission, not just a compromise, so the "bridge-attested destination" mitigation is weaker than the NatSpec states.
- **`allowUnverified` env read (:582)** — the only deploy input not read in `run()` into `DeploymentContext`; that's what forces `vm.setEnv` in the tests, and a revert between set and restore leaves the escape hatch on for the rest of the process.

scripts/Deploy.s.sol
carlosfebres: Fixed in 98659d90. Reproduced it first — a member declaring `chainIdByDomain(uint64) returns (uint256)` and returning `type(uint256).max` makes the probe revert with no data, which is exactly the uncatchable frame you describe.

Worth noting it was worse than the lane loop: this probe also runs unconditionally at `:623` with domain 0 as the bridge-attestation gate, so every member hit it, not just ones with a configured lane.

Now decodes wide and range-checks, per your snippet. The `@dev` that claimed the length check "avoids that trap entirely" is corrected rather than left standing. Pinned by `test_rejectsMemberWithDirtyChainIdByDomain` with a new `MockDomainProverDirtyChainId`; I confirmed the pin bites by reverting the fix — it fails with "call reverted as expected, but without data" instead of the intended require string.

scripts/Deploy.s.sol
carlosfebres: Fixed in e7cd8808, taking the first of your two options. A member matching `ctx.hyperProver`/`ctx.metaProver` now requires a non-empty domain config; every configured lane still round-trips as before.

`LayerZeroProver` is deliberately exempt, for the reason you gave — strict map, no fallback, reverts `UnregisteredDomain`, so an omitted lane yields no proof rather than a wrong destination. `test_acceptsEmptyDomainConfigForStrictMapMember` pins the exemption so the requirement is not silently widened later.

To be explicit about the blast radius: this applies only when the prover is an aggregator **member**. `HYPER_DOMAIN_CONFIG`/`META_DOMAIN_CONFIG` stay exceptions-only for an ordinary deploy. The NatSpec on `provenIntents` and the validator `@dev` are rewritten — they now state what is actually enforced instead of implying bridge attestation covers the domain map unconditionally.

scripts/Deploy.s.sol
carlosfebres: Fixed in 2200e61c. `AGGREGATOR_PROVER_ALLOW_UNVERIFIED_MEMBERS` is read once in `run()` into `ctx.allowUnverifiedMembers`, so `validateAggregatorProverMembers` is a pure function of the context.

The two merged cases are independent tests again (`test_rejectsUnknownMemberWithoutEscapeHatch` / `test_acceptsUnknownMemberWithEscapeHatch`), each setting the struct field on its own ctx. No `vm.setEnv` anywhere in the file now, so the revert-skips-the-restore hazard you describe is gone rather than worked around.

contracts/prover/AggregatorProver.sol
carlosfebres: Taken in 2d57ab6e — decodes to `(bytes32, uint256)` and uses `rawClaimant.isValidAddress()` then `.toAddress()`.

Same semantics (`isValidAddress` is the same `>> 160` test), but it keeps one canonical notion of "this word is an address" and makes the `toAddress()` provably revert-free, since the check precedes it.

contracts/prover/AggregatorProver.sol
carlosfebres: Done in de5fd0fe — the constructor now reverts `MemberHasNoCode(address)` and both runtime `code.length` guards are removed.

Two things I checked before removing them, since the read paths differ:

- `provenIntents` uses a low-level `staticcall`, and a codeless target returns success with **empty** returndata, so the existing `ret.length != 64` check already covers it. Removing the guard there is safe and saves an `EXTCODESIZE` per member.
- `challengeIntentProof` is a high-level void `try` call, where solc emits its own `extcodesize` check that does **not** land in the `catch`. I confirmed that empirically — a `try` against a codeless address reverts outright with neither branch running. So there the guard was doing real work and its removal rests entirely on the constructor invariant.

That invariant holds because code cannot disappear post-EIP-6780, and there is no `AggregatorProverTron` variant (TRON deploys via `scripts/tron`), so no target chain predates Cancun.

Accepted tradeoff, called out in the NatSpec and `scripts/README.md`: deploying on a chain where a member is not live yet now **reverts** instead of silently deploying a smaller trust set. That is the fail-closed direction, but it does interact with re1ro's §5 point about chains onboarded mid-release.

contracts/prover/AggregatorProver.sol
carlosfebres: Pushing back on this one, or at least narrowing it.

Line 229 converts the value returned by `getWhitelist()`, which is `bytes32[]` because it comes from the `Whitelist` base that provides the immutable packed storage. So changing the constructor to `address[]` would not remove that conversion — the read path still gets `bytes32` words back. It would only change the external deploy ABI, and it would diverge from every other prover, which all take `bytes32[] _proverAddresses` for cross-VM compatibility.

Eliminating the conversion for real means not using `Whitelist` for storage, which is a much larger refactor than this line implies.

What I did take from it is the type-safety half, via your other comment: the read path now decodes to `bytes32` and goes through `AddressConverter.isValidAddress`/`toAddress`, so the conversion is checked by the shared library rather than an inline shift. Happy to revisit the constructor signature separately if you still want it.

carlosfebres:
@re1ro — responses by section. The work is now split one commit per issue rather than one squashed commit; per-section SHAs below.

**§3 — probe waves through the empty-dynamic class.** Fixed in `3c98bb32`, using your one-liner. `_tryProvenIntentsShape` now requires both words to be exactly zero, which an honest member satisfies because `bytes32(0)` is unproven. Pinned by `test_rejectsMemberWithEmptyDynamicProvenIntents` against a new `MockDomainProverEmptyDynamic`; I verified the pin bites by reverting the fix. Thanks for having already run it against `56f8ea2` — matches what I see.

**§4 / §5 — validation disabled after the first deploy, and per-chain sets discarded.** Took the salt fix, in `1c4986f9`. `getAggregatorProverSalt` = root salt + name + `keccak256(abi.encode(members))`, so a member change re-derives the aggregator address on its own without touching `SALT` or any other prover. Order is inside the hash deliberately, since order is priority — `[Hyper, Meta]` and `[Meta, Hyper]` must not collide. Four tests cover set-change, order-change, non-collision with the plain name salt, and determinism.

The consequence you flagged in §5 is now explicit rather than accidental: the aggregator only shares one address across chains whose member sets are identical. That seems like the honest outcome given the member set is a constructor arg. The `deployedAll.csv` cross-chain comparison you describe is **not** done — the data is still recorded and discarded. Worth a follow-up ticket unless you think the salt change covers enough of it.

**§6 — gas.** Measured and pinned in `7ea1ffd6` so it is written down: worst case at `MAX_MEMBERS` with no member holding the proof is **31,799 gas cold / 15,762 warm** (`test_gas_worstCaseFanOutAtMaxMembers`, which logs both and has regression tripwires). Cold is the number to weigh against your 150k blind-fallback floor. Lower than your +49k, which I assume covered more of the withdraw path than the read alone.

**§1 and §2 — solver-side.** Confirmed with @cfebres: you own §1, resolving `sourceChainProver` per prover type rather than from `reward.prover`. We are deliberately **not** adding the belt-and-braces send-path guard, because it would modify `HyperProver`/`MetaProver`/`LayerZeroProver`, and under this repo's policy a change to possibly-deployed contracts cannot go out as a normal public PR. If you want that guard anyway, it needs the deployed-status confirmation first and would go through the private advisory path.

Your §2 write-up is the most useful thing in this review for rollout — particularly that adding `aggregator` to `ProverTypeValues` alone would half-wire it into quoting while `getProver()` still returns `null`. Agreed a `BaseProver` subclass is mandatory, and that the **Pre-deployment gate** in the PR description should name solver support. Updating that now.

**Nits.** The reentrancy row is in the PR description, not the contract — scoping it to `provenIntents`, since `challengeIntentProof` is a plain `CALL`. The already-deployed member-set guard now has a test pinning the property it depends on (`getWhitelist()` returns members in exact constructor order, left-padded). PR description is being refreshed to match the current branch.

Finally, one change lands directly on your §5 onboarding case — members must now have code at construction, so a chain onboarded before all members are live fails the deploy rather than producing a smaller set. Details in [the `:311` thread](https://github.com/eco/eco-routes/pull/433#discussion_r3847885397); flagging here only because it may affect your rollout sequencing.

contracts/prover/AggregatorProver.sol
fish-sammy: It's not about removing the conversion from the entire contract but about removing it from the hot path. `getWhitelist()` is a getter, gas-free when being queried via RPC. The constructor is a one-time thing. However, `prove` is the hot path, and removing the conversion from there saves gas. You can still use the `Whitelist` by calling `AddressConverter.toBytes32` in the constructor.

fish-sammy (approved):

re1ro (approved):
Approving. I re-verified rather than reading the commit titles — checked out `7ea1ffd6`, ran the suites, and tried to falsify the new pins.

**Verified myself:**

- **§3 probe.** `_tryProvenIntentsShape` is now `rawClaimant == 0 && rawDestination == 0` (Deploy.s.sol:593). Reverted it to the old predicate and `test_rejectsMemberWithEmptyDynamicProvenIntents` fails — the pin bites. Worth noting what the failure revealed: with the fix reverted the member is still rejected, just by the *domain-config* check instead, so there was more defence in depth than my §3 implied. The fix is still right — it rejects for the correct reason, and the domain check is conditional on `needsExplicitLanes`.
- **§4/§5 salt.** `getAggregatorProverSalt = keccak256(getContractSalt(root, "AGGREGATOR_PROVER"), keccak256(abi.encode(members)))`, with order inside the hash. Four tests present as described: set-change, order-change, non-collision with the plain name salt, determinism.
- **Gas.** `31,799 cold / 15,762 warm` at `MAX_MEMBERS`, logged by the test. You're right that my +49k covered more of the withdraw path than the read alone — yours is the number to carry.
- **Suites.** 26 in `AggregatorProverMemberValidation`, 63 across the aggregator contracts. All green.

**Two things you found that I missed, both better than what I raised:**

`de5fd0fe` — a codeless member would revert `challengeIntentProof` *outright*, because solc's extcodesize check on a void high-level call lands in the caller's frame, outside the catch. That breaks the never-reverts guarantee, and I didn't spot it. The EIP-6780 reasoning for checking once at construction is sound: post-Cancun, code can't disappear after creation. Failing the deploy rather than silently deploying a smaller set is the right direction.

`e7cd8808` — **I had §4 backwards.** I wrote that the real loss was on LayerZero members "whose strict domain→chainId map has no fallback." That's the safe case. Confirmed in the source: `HyperProver.sol:106-107` and `MetaProver.sol:113-114` both do `chainId != 0 ? chainId : originDomain`, so an omitted lane silently records a plausible-looking *wrong* chain id, while LayerZero's strict map fails closed. Requiring explicit lanes for the fallback resolvers is the correct inversion of what I said.

**On what you didn't take:**

- **§1 send-path guard** — declining because a change to possibly-deployed contracts can't go out as a normal public PR is a real reason, not a brush-off. Agreed it belongs on the private advisory path if we want it. I own the solver half regardless: resolving `sourceChainProver` per prover type rather than from `reward.prover`. That work is mine and doesn't block this.
- **§5 cross-chain `deployedAll.csv` comparison** — agreed, follow-up ticket. The salt change does most of the work: differing member sets now derive *different* addresses, so the specific hazard I described (same address, different sets, nothing compares) is largely gone. What's left is a reporting nicety.

Good call splitting one commit per issue — it made this re-review a delta rather than a re-read.

Please make sure the **Pre-deployment gate** names solver support before an aggregator address is published anywhere the solver serves; per §2 that's still the thing standing between this and an intent nobody can fulfill.

PR #438, opened 2026-08-27
feat(chain): add IntentChainer for runtime-amount intent chaining

re1ro:
Closes [PAR-586](https://linear.app/ecoprotocol/issue/PAR-586/intentchainer-publish-a-follow-on-intent-whose-amount-is-only-known-at)

- [x] This PR is not a security fix. It adds a feature and explicitly changes the prefunding policy.

V4 was deployed at the operator's request. Current V5 source is not deployed.
Latest review follow-up: [4ed1aa1](https://github.com/eco/eco-routes/commit/4ed1aa148492a91f7ada5a25b20a35ddccf40d09).

## Problem

A runtime amount determines an intent's reward hash, intent hash, and vault address.
Amount-only slots cannot template a later CCTP mintRecipient for a remote, dynamically
constructed intent. The vault locally funded by chain() and the remote recipient named
inside its generated route are separate objects.

Review also requested a clearer child-identity boundary, separated output calculation,
and end-to-end outbound short-transfer rollback coverage.

## Impact

Enables dynamic EVM/TRON vaults and Solana vault ATAs inside the child route without
additional escrow. Derivation errors can misdirect a later mint, so configs, encodings,
graph references and canonical PDA selection fail closed.

V5 intentionally permits existing local child-vault balances. Builders own child-salt
uniqueness; a top-up does not increase a child's declared reward or delivery obligation.

## What Changed

- Dependency-first downstream vault nodes and root templates containing amount/vault items.
- Tagged ABI configs; explicit EVM/TRON deployment parameters; canonical Solana vault PDA
  and ATA derivation using SHA-256, Edwards25519 membership checks and MODEXP.
- Removed VaultAlreadyFunded at the author's request. Minimum amounts, terminal-state
  rejection, Inbox parent replay checks and transfer-delta verification remain.
- Extracted `_calculateAmountOut`; zero/codeless local Portals fail with InvalidPortal.
- Added real Portal/Executor fee-on-transfer rollback tests and mined-receipt event tests.
- Updated NatSpec/README for optional publication, builder-owned salts, opaque route
  deadlines, donations and bounded concatenation.
- Deployment scripts select INTENT_CHAINER_V5; the V4 manifest remains unchanged.

## Human Logic

```text
validate local Portal and order
measure amountIn and enforce minimum; calculate ceil-rounded amountOut
for each downstream vault, dependency-first:
    render its complete route and VM-native reward
    hash destination, routeHash and rewardHash
    derive the explicit EVM/TRON vault or canonical Solana vault ATA
render root route and overwrite local reward amount with amountIn
resolve LOCAL child hash/vault and reject terminal state
optionally publish the child
read vault balance; transfer amountIn; require the balance delta to cover amountIn
emit IntentChained

existing vault balance is allowed and not counted toward the new transfer
builders choose fresh child salts when independent children are required
later: another fulfillment executes the child's already-populated route
```

## Logic Trace

```mermaid
sequenceDiagram
    participant X as Parent Executor
    participant C as IntentChainer
    participant P as Local Portal
    participant V as Child vault
    participant Y as Later child Executor
    participant B as Bridge
    X->>C: Token output, then committed Order
    C->>C: Measure, calculate, render and derive recipients
    C->>P: Resolve local vault; reject terminal state
    opt publish
        C->>P: Publish completed child
    end
    C->>V: Read balance; push measured amount; verify delta
    Note over C,V: Prefunding is allowed
    Note over X,V: Failure reverts parent transaction and logs
    Y->>B: Later populated burn amount and mintRecipient
```

## Service Topology

```mermaid
flowchart LR
    X[Parent Executor] -->|committed Order and tokens| C[IntentChainer]
    C -->|render only| T[Template renderer]
    T -->|explicit parameters| E[EVM/TRON formula]
    T -->|canonical PDA and ATA| S[SHA-256 and MODEXP]
    C -->|optional publish and status| P[Local Portal]
    C -->|local push and balance delta| V[Child reward vault]
    T -. completed route .-> Y[Later child Executor]
    Y -->|later burn| B[CCTP]
    B -->|later mint| R[Remote vault or Solana ATA]
```

## Why This Shape

- Both route and reward contribute to the downstream hash; both need templates.
- Dependency-first references support nesting without cycles or self-references.
- Remote recipients are route data, not replacement local funding targets. No additional
  escrow, approvals, callbacks or chainer-side burn is introduced.
- Canonical bump computation supersedes the original supplied-bump proposal: fixed
  program IDs do not fix bumps when amounts change. [Sauce PR #416](https://github.com/eco/sauce/pull/416)
  demonstrates the EVM field-arithmetic/MODEXP approach.
- Child identity is caller-owned. Distinct parents may reuse a child; Inbox still rejects
  replay of the same parent. Existing vault balance is not a uniqueness check. Transfer
  delta independently verifies new delivery, regardless of prefunding.
- CREATE3 ignores bytecode. V5 needs a new salt because deployed V4 is immutable; V5
  retains V4's Order/chain ABI but deliberately changes the prefunding policy.

## Boundaries

- At most 8 downstream nodes, 8 items per template and 32 KiB total rendered bytes.
- Each Solana PDA search tries bumps 255 through 224 and fails closed on exhaustion;
  caller bumps are not accepted. Precompiles must have standard output and behavior.
- Builders verify remote deployment parameters, VM-native encodings, token/prover IDs,
  unique child salts, route deadlines and bridge fees. Only the local reward deadline
  receives the on-chain five-minute buffer; opaque route deadlines are not checked here.
- Reusing the complete child preimage and amount funds the same unsettled vault. The
  declared reward is paid once; surplus can be refunded to the creator after withdrawal.
- Amount transforms remain proportional. No flat-fee expression was added.
- Any caller-selected token donated to the shared chainer remains sweepable by the next
  caller. The intended atomic flow leaves no balance at rest; this exposure is unchanged.
- V4 remains deployed at `0x96f7DBa7E587f359a8f585C335fe510E4dC9bCC0` on chains 1, 10, 56,
  130, 137, 143, 146, 480, 999, 8453, 9745, 42161, 42220 and 57073. Its
  [manifest](https://github.com/eco/eco-routes/blob/feat/intent-chainer/deployments/intent-chainer-v4.json)
  is historical V4 evidence, not a V5 rollout record.
- V5 is undeployed. Adoption needs a chainer-only rollout on execution chains and client
  address updates. Default rollout remains the 14 chains above; excluded unfunded chains
  remain 169, 466, 2020, 5000, 5330, 8333, 33139 and 10241024.
- No Portal/Vault replacement, Solana Portal upgrade, solver/SDK rollout, asset transfer
  or bridge operation is included. A remote recipient alone does not need a chainer there.
  Revalidate HyperEVM deployment gas headroom for V5.

## Validation

| Gate | Result |
| --- | --- |
| Full Foundry | 819 passed, 0 failed, 0 skipped; 43 suites |
| Hardhat / Jest | 114 Hardhat and 9 Jest tests passed |
| Short-transfer rollback | Exact wrapped PushShortfall; solver, swapper, chainer, vault, total supply and claimant rollback |
| Mined receipts | 2 failures, publish on/off: status 0, empty receipt logs, no surviving published/chained events |
| Prefunding | Below/equal/above-push balances, distinct-parent child reuse, same-parent replay rejection, one reward and surplus refund |
| Amounts and derivation | Existing scaling, independent EVM/TRON and SVM fixtures, cross-VM and malformed-input tests retained |
| Deployment | 12 script tests including V5 salt selection; V4 manifest unchanged; no V5 broadcast |
| Static | Prettier, Solidity lint, explicit TypeScript ESLint, shell syntax and whitespace checks pass |

Solana fixtures use the independent Solana SDK and eco-routes-svm ref
`95c728850955d3c9aa7c62ba08a87db5be89e220`, including the Portal's vault golden; they are
not generated by the Solidity helper.

### DIFF
--- a/contracts/chain/IntentChainer.sol
+++ b/contracts/chain/IntentChainer.sol
@@ -17,10 +17,10 @@ import {Reward} from "../types/Intent.sol";
  *      output of whatever intent1's earlier calls produced -- typically a DEX swap that named this contract
  *      as its recipient. It measures that balance -- the ONE runtime measurement in the whole flow -- and
  *      derives from it the two amounts of a second intent (intent2) whose every other field was fixed when
- *      intent1 was signed: `amountIn` is escrowed as intent2's reward, and the fee-deducted, unit-scaled
- *      remainder is written into the route as what intent2's solver must deliver. The flat `fee` is that
- *      solver's entire margin; the scale converts source units into destination units, which are NOT the
- *      same thing for the same token across chains.
+ *      intent1 was signed: `amountIn` is escrowed as intent2's reward, and `amountIn * scale` is written
+ *      into the route as what intent2's solver must deliver. The gap between them is that solver's margin,
+ *      and `scale` also carries the source-to-destination unit conversion -- which is NOT the identity for
+ *      the same token across chains.
  *
  *      WHY THIS IS SAFE WITHOUT ACCESS CONTROL. Route calls reach every contract through the Portal's single
  *      shared {Executor}, which any fulfiller of any intent drives permissionlessly, so `msg.sender` proves
@@ -51,7 +51,7 @@ contract IntentChainer is ReentrancyGuard {
 
     /**
      * @notice A position in intent2's route bytes that receives the destination amount.
-     * @dev Every slot receives the SAME scalar, `ceil((amountIn - fee) * scale / WAD)`. That is what the
+     * @dev Every slot receives the SAME scalar, `ceil(amountIn * scale / WAD)`. That is what the
      *      solver of intent2 must supply on the destination, and it appears once as
      *      `route.tokens[0].amount` and again inside any
      *      call that moves it (the `transfer(swapper, ...)` that feeds a destination swap). Because both
@@ -80,29 +80,32 @@ contract IntentChainer is ReentrancyGuard {
      * @param slots Positions receiving the measured amount, in route order.
      * @param reward Intent2's reward. MUST carry exactly one leg, in `token`, with no native amount; the
      *        leg's `amount` field is ignored on input and overwritten with the measured amount.
-     * @param fee Flat spread kept by intent2's solver, in the SOURCE token's smallest unit, deducted
-     *        BEFORE scaling. The reward leg escrows `amountIn` while the route obliges the solver to
-     *        deliver the scaled remainder, so the difference is its entire margin. Fixed at intent1
-     *        signing time, so it is a flat fee on a floating amount, not a rate -- a larger-than-expected
-     *        swap output raises what the solver must deliver one-for-one and leaves its take unchanged.
-     * @param scale Source-to-destination conversion, {WAD}-denominated:
-     *        `amountOut = ceil((amountIn - fee) * scale / WAD)`.
+     * @param scale The ENTIRE source-to-destination transform, {WAD}-denominated:
+     *        `amountOut = ceil(amountIn * scale / WAD)`. It carries both the unit conversion and intent2's
+     *        solver spread, because those compose into one ratio and there is no reason to commit two
+     *        numbers where one will do.
      *
-     *        The conversion exists because "the same token" does not mean the same UNIT across chains.
-     *        USDC is 6 decimals on Ethereum, Base and Solana, but Binance-Peg USDC on BNB Chain is 18, and
-     *        Arc's native USDC is 18 against a 6-decimal ERC20 wrapper (see
-     *        {DepositAddress_CCTPMint_Arc}'s `NATIVE_USDC_SCALING`).
+     *        THE UNIT PART exists because "the same token" does not mean the same UNIT across chains. USDC
+     *        is 6 decimals on Ethereum, Base and Solana, but Binance-Peg USDC on BNB Chain is 18, and Arc's
+     *        native USDC is 18 against a 6-decimal ERC20 wrapper (see {DepositAddress_CCTPMint_Arc}'s
+     *        `NATIVE_USDC_SCALING`).
      *
      *        The denominator is DECIMAL on purpose. Unit conversions are powers of ten, so a decimal
      *        denominator represents every one of them exactly in both directions, where a binary
-     *        denominator (Q128 and friends) cannot: 6-to-18 is `scale = 1e30`, 18-to-6 is `scale = 1e6`,
-     *        and a same-unit lane is `scale = WAD`, all integers.
+     *        denominator (Q128 and friends) cannot: 6-to-18 is `1e30`, 18-to-6 is `1e6`, and a same-unit
+     *        lane is `WAD`, all integers. A spread multiplies in: same units less 50bps is `0.995e18`.
      *
-     *        The ratio is applied proportionally, so it carries a price as well as a unit change without
-     *        reintroducing a surplus leak: a bigger `amountIn` raises the obligation in the same
-     *        proportion. Rounding is toward the USER (up), because this value is the solver's delivery
-     *        FLOOR -- rounding it down would quietly shave the last unit off what the user receives on
-     *        every downscaling lane.
+     *        THE SPREAD PART is PROPORTIONAL, not flat. The reward leg escrows the whole measured
+     *        `amountIn` while the route obliges only `amountIn * scale`, so the difference is the solver's
+     *        entire margin, and it grows with the amount. That is a deliberate trade: a flat fee would
+     *        price destination gas independently of size, but it cannot be folded into a ratio, and the
+     *        only thing that moves `amountIn` here is swap slippage -- a percent or so around a known
+     *        expectation -- over which the two are indistinguishable. Use `minAmountIn` to express "too
+     *        small to be worth filling"; it says that directly, where a flat fee only said it by accident.
+     *
+     *        Rounding is toward the USER (up), because this value is the solver's delivery FLOOR --
+     *        rounding it down would quietly shave the last unit off what the user receives on every
+     *        downscaling lane.
      * @param minAmountIn Floor on the measured amount. Reverts below it, so a swap that under-delivered
      *        unwinds intent1 rather than publishing an intent nobody will fill.
      */
@@ -112,7 +115,6 @@ contract IntentChainer is ReentrancyGuard {
         bytes[] segments;
         Slot[] slots;
         Reward reward;
-        uint256 fee;
         uint256 scale;
         uint256 minAmountIn;
     }
@@ -182,9 +184,6 @@ contract IntentChainer is ReentrancyGuard {
 
     /// @notice The measured amount is below the order's floor.
     error AmountBelowFloor(uint256 amountIn, uint256 minAmountIn);
-
-    /// @notice The flat fee is at or above the measured amount, leaving nothing to deliver.
-    error FeeExceedsAmount(uint256 amountIn, uint256 fee);
 
     /// @notice {Order.scale} is zero, which would oblige the solver to deliver nothing.
     error InvalidScale();
@@ -271,13 +270,10 @@ contract IntentChainer is ReentrancyGuard {
 
     /**
      * @notice Measure the held balance and derive the two amounts intent2 is built from.
-     * @dev The fee is taken FIRST, in source units, and only then converted to destination units. The other
-     *      order would make the solver's take depend on the scale factor rather than on the fee alone.
-     *
-     *      Ceil rounding serves two purposes. It rounds toward the USER, since `amountOut` is the solver's
-     *      delivery floor. And it guarantees a non-zero obligation: with `netAmountIn >= 1` and
-     *      `scale >= 1` the quotient is strictly positive, so no order can oblige a solver to deliver
-     *      nothing while collecting the whole escrow. That is why there is no zero-obligation check.
+     * @dev Ceil rounding serves two purposes. It rounds toward the USER, since `amountOut` is the solver's
+     *      delivery floor. And it guarantees a non-zero obligation: with `amountIn >= 1` and `scale >= 1`
+     *      the quotient is strictly positive, so no order can oblige a solver to deliver nothing while
+     *      collecting the whole escrow. That is why there is no zero-obligation check.
      * @param order The validated order.
      * @return amountIn The measured balance, escrowed as intent2's reward.
      * @return amountOut The scaled destination obligation.
@@ -293,16 +289,8 @@ contract IntentChainer is ReentrancyGuard {
         if (amountIn < order.minAmountIn) {
             revert AmountBelowFloor(amountIn, order.minAmountIn);
         }
-        if (amountIn <= order.fee) {
-            revert FeeExceedsAmount(amountIn, order.fee);
-        }
 
-        amountOut = Math.mulDiv(
-            amountIn - order.fee,
-            order.scale,
-            WAD,
-            Math.Rounding.Ceil
-        );
+        amountOut = Math.mulDiv(amountIn, order.scale, WAD, Math.Rounding.Ceil);
     }
 
     /**
--- a/contracts/chain/IntentChainer.sol
+++ b/contracts/chain/IntentChainer.sol
@@ -1,0 +1,347 @@
+/* -*- c-basic-offset: 4 -*- */
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.26;
+
+import {IERC20} from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import {SafeERC20} from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+import {ReentrancyGuard} from "@openzeppelin/contracts/utils/ReentrancyGuard.sol";
+
+import {IIntentSource} from "../interfaces/IIntentSource.sol";
+import {Reward} from "../types/Intent.sol";
+
+/**
+ * @title IntentChainer
+ * @notice Publishes a second intent whose amount is not known until the first intent executes.
+ * @dev Called as the last {Call} of an executing intent (intent1), at which point this contract holds the
+ *      output of whatever intent1's earlier calls produced -- typically a DEX swap that named this contract
+ *      as its recipient. It measures that balance -- the ONE runtime measurement in the whole flow -- and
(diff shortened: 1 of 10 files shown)


re1ro (commented):
## Review — contracts/chain/IntentChainer.sol

Read at `e354167` against `contracts/{IntentSource,Inbox,Executor}.sol`, `interfaces/IIntentSource.sol` and the three chain tests, plus `contracts/deposit/*` for the encoder the Borsh fixture leans on.

Reciprocating the review on eco-routes-svm#89 — and finding 1 is the mirror image of the one you raised there, which turns out to apply here too.

The `Math.mulDiv(..., Rounding.Ceil)` choice, the segments representation, and the `_validate`-before-`_measure`-before-`publish` ordering are all right, and the Borsh fixture's non-circularity is the strongest single test in either repo.

---

### 1. The salt-collision revert does not fire in the window that matters — medium

The design rests on `publish` rejecting a colliding hash. `_validatePublish` (`IntentSource.sol:850-856`) rejects only two of four states:

```solidity
if (status == Status.Withdrawn || status == Status.Refunded) {
    revert IntentAlreadyExists(intentHash);
}
```

`Initial` and `Funded` both pass — and the comment there says so deliberately, "publishing is idempotent and repeatable rather than once-per-intent."

But this contract never calls `fund`, by design, so **intent2 sits at `Status.Initial` for its entire useful life** — the README states exactly that at line 165. So for the whole window between chaining and settlement, a second `chain` on a colliding order does *not* revert. It publishes again and pushes a second `amountIn` into the same vault.

`PushShortfall` cannot catch it, because `delivered` is an **absolute** balance (`IntentChainer.sol:261`): after the second push the vault holds `2 * amountIn`, and `2 * amountIn >= amountIn` passes.

`test_atomicity_settledIntentTwoHashRevertsBeforeAnyPush` withdraws intent2 first and asserts `Status.Withdrawn` before collidingdirectly — so it pins the terminal case and leaves the `Initial` case, which is the normal one, uncovered.

Concretely: chain an order, don't settle it, refill the chainer, chain the identical order again. Second push lands. One intent2 exists; the claimant withdraws the reward leg once; the surplus sits until `reward.deadline` and refunds to `reward.creator`. Not a loss when the second chainer *is* the creator, but a silent merge with no signal — the second intent1 was consumed and produced no second delivery.

Worth noting this inverts the asymmetry in your review of #89: you wrote "on EVM the same collision reverts." It reverts only after settlement. The window where it doesn't is the operating window.

**Fix** is the one I took on the SVM side, and you already have the read: check the vault *before* pushing, not only after.

```solidity
if (IERC20(order.token).balanceOf(vault) >= amountIn) revert VaultAlreadyFunded(vault);
```

Nothing legitimate pre-funds intent2's vault — the address depends on the measured amount — so it has no false-positive surface, and a dust donation doesn't trip it.

### 2. `PushShortfall` compares an absolute balance, so a donation masks a fee-on-transfer — low

Same line. `delivered = balanceOf(vault)` measures the vault's total, not what this transfer actually moved. Any pre-existing balance covers a shortfall.

The order is public in intent1's calldata, so the vault address is computable by anyone watching the mempool, and a donation of the fee amount would mask exactly the case this guard exists to catch. There's no profit in it, so it's low — but a delta is strictly more correct and costs one extra `balanceOf`:

```solidity
uint256 before = IERC20(order.token).balanceOf(vault);
IERC20(order.token).safeTransfer(vault, amountIn);
if (IERC20(order.token).balanceOf(vault) - before < amountIn) revert PushShortfall(...);
```

Folding fix 1 into this gives both for one read.

### 3. Intent2's *route* deadline is neither checked nor checkable — low, doc-level

`MIN_DEADLINE_BUFFER` guards `reward.deadline`. The route deadline lives inside the opaque bytes and can't be read here, so an intent2 whose route deadline has already passed publishes and funds successfully, is unfulfillable, and locks the escrow until `reward.deadline` — the longer of the two by construction.

Same limitation exists on my side and neither of us can close it in the contract. The "Deadline headroom" SDK item covers the reward deadline; it's worth saying explicitly that the route deadline is the one with no on-chain backstop at all.

### Minor

- `_buildRoute` calls `bytes.concat` in a loop, reallocating and copying the whole accumulated route each iteration — quadratic in segment count. Bounded by `MAX_SLOTS = 8` so it's ≤8 copies and fine today; worth a comment that the bound is what keeps it acceptable, since the cost is invisible at the call site.
- `order.token` is fully caller-chosen and `chain` sweeps that token's entire balance, so the documented donation risk extends to *any* token mistakenly sent to a widely-known singleton — not just the lane's own. Accepting it is reasonable; the README's "tokens donated to the chainer" row could say "any token", since that's the part people will be surprised by.

### Verified rather than flagged

- **`Math.mulDiv(amountIn, order.scale, WAD, Rounding.Ceil)`** — the 512-bit intermediate makes the overflow that forced a gcd reduction on my side a non-issue here, and ceil-toward-user plus zero-unreachable both hold. Confirmed the same lanes agree with my implementation byte-for-byte via the shared vector.
- **`_encodeAmount`** — correct at widths 8/16/32 in both orders; `width < 32` guards the shift, and 32 needs no check since a `uint256` always fits.
- **Non-payable with no `receive()`** — checked `Refund.excessNative()` reached from every funding entry point. The reasoning holds and is worth keeping verbatim in the NatSpec.
- **`_validate` before `_measure` before `publish` before transfer** — no state or value moves ahead of any revert path.

### Follow-up

`testdata/cross-vm-vectors.json` is in eco-routes-svm#89 and asserted by four tests there. Pulling it in here and asserting from Foundry pins parity in both directions — happy to add an upscaling lane (`scale = 1e30`) once it reads, since that's where the two implementations diverge most.

re1ro:
## Findings 1 and 2 fixed — `4174626`

Both real, and finding 1 is a genuine bug I had the wrong model of. Thanks for checking `_validatePublish` against the `Initial` state rather than taking the PR body's word for it — that's exactly where I was wrong.

### 1 + 2 — collapsed into one bracket around the transfer

You were right that they fold together, and the ordering is the whole fix:

```solidity
uint256 balanceBefore = IERC20(token).balanceOf(vault);
if (balanceBefore >= amountIn) revert VaultAlreadyFunded(vault, balanceBefore);

IERC20(token).safeTransfer(vault, amountIn);

uint256 delivered = IERC20(token).balanceOf(vault) - balanceBefore;
if (delivered < amountIn) revert PushShortfall(vault, amountIn, delivered);
```

The framing that convinced me is the one you led with: `publish` is idempotent on `Initial` **by design**, and this contract never calls `fund`, so `Initial` isn't an edge state — it's the only state intent2 is ever in before settlement. The rejection I was relying on covers the half that can't happen yet.

The test was the tell. `test_atomicity_settledIntentTwoHashRevertsBeforeAnyPush` withdraws first and asserts `Withdrawn` before colliding, so it pinned the terminal case and left the normal one uncovered — which is precisely how I convinced myself the property held. Added `test_atomicity_collidingHashRevertsWhileIntentTwoIsUnsettled`, and **verified it fails without the guard**: without the balance read the second `portal.fulfill` doesn't revert at all. Both halves are now pinned.

### 3 — route deadline

Agreed, and documented rather than fixed since neither of us can close it in a contract. The README now separates it from the reward-deadline item and says explicitly that it is the one deadline with **no on-chain backstop on either VM**.

### Minor — both taken

`bytes.concat` in a loop now carries a comment saying `MAX_SLOTS` is what makes the quadratic cost acceptable, and that raising the bound means revisiting the loop rather than just the constant. The donation row in the README now reads "any token sent to the chainer", with a note that `order.token` is order-chosen and the sweep takes that token's whole balance — you're right that the surprise is the generality, not the mechanism.

### Cross-VM vectors — imported and binding

`testdata/cross-vm-vectors.json` is vendored here verbatim and asserted by `IntentChainerCrossVmTest`, which drives the real `chain` entry point and compares the published route byte-for-byte rather than testing an internal helper. So the fixture now binds in both directions instead of one.

Please do add the `scale = 1e30` upscaling lane — that's where your gcd reduction and my `mulDiv` 512-bit intermediate take genuinely different paths to the same number, so it's the vector most likely to catch a future divergence. I'll confirm it reproduces here as soon as it lands.

750 forge tests pass, 0 failures.

MSevey (changes requested):
Matthew review complete — REQUEST CHANGES.

BLAST RADIUS: Tier 0 — this new on-chain contract measures an ambient token balance during intent1 fulfillment, constructs intent2, and transfers that balance into the deterministic reward vault. A short push can directly underfund a proven reward.

MUST-FIX

1. Add a real outbound fee-on-transfer regression for `_pushAndVerify` in `contracts/chain/IntentChainer.sol:350-366`. The balance-delta guard is implemented specifically for deflationary tokens, but no test executes that branch; `test_atomicity_underDeliveredSwapRevertsTheWholeFulfillment` only exercises the inbound `minAmountIn` floor. Use a fee-on-transfer ERC20 through the real `Portal.fulfill` path, assert the exact `PushShortfall`, and prove whole-transaction rollback: the chainer retains its pre-push balance, the vault remains unchanged, intent1 is not recorded fulfilled, and no publish/chained event survives. Missing failure-mode coverage on a Tier 0 vault-funding path blocks landing.

CAN-SLIDE (non-blocking)

- `IntentChainer.sol:40-45` and `:241-244` still say publish is unconditional/returns the vault, contradicting the current optional-publish and `intentVaultAddress` implementation.
- `InvalidPortal` is declared but never used; a zero/codeless Portal currently fails later with an opaque ABI-decode revert.
- `deployIntentChainers.sh:11` still names `INTENT_CHAINER_V1`, while the actual salt discriminator is V3.

RECOMMENDATION: ship after the short-transfer rollback invariant is pinned.

contracts/chain/IntentChainer.sol
```diff
@@ -0,0 +1,369 @@
...
+        amountIn = IERC20(order.token).balanceOf(address(this));
+
+        if (amountIn == 0) {
+            revert ZeroAmount();
+        }
+        if (amountIn < order.minAmountIn) {
+            revert AmountBelowFloor(amountIn, order.minAmountIn);
+        }
+
+        amountOut = Math.mulDiv(amountIn, order.scale, WAD, Math.Rounding.Ceil);
```
Foivos, line 327: I would maybe put this in a separate function in case we want to do fancier stuff with it later (like adding a flat fee).

contracts/chain/IntentChainer.sol
```diff
@@ -0,0 +1,369 @@
...
+     * @param vault Intent2's vault.
+     * @param amountIn The amount to push.
+     */
+    function _pushAndVerify(
+        address token,
+        address vault,
+        uint256 amountIn
+    ) internal {
+        uint256 balanceBefore = IERC20(token).balanceOf(vault);
+        if (balanceBefore >= amountIn) {
```
Foivos, line 280: This is confusing to me. If an intent (and its vault) could have been funded before shouldn't we check that the balance is 0 instead?

contracts/chain/IntentChainer.sol
re1ro: Addressed in [4ed1aa1](https://github.com/eco/eco-routes/commit/4ed1aa148492a91f7ada5a25b20a35ddccf40d09): extracted `_calculateAmountOut(amountIn, scale)` from `_measure`. Measurement and minimum checks stay in `_measure`; the proportional formula and ceiling rounding are unchanged. No flat fee was introduced. Existing scaling, rounding and cross-VM fixture tests pass.

contracts/chain/IntentChainer.sol
re1ro: Removed `VaultAlreadyFunded` in [4ed1aa1](https://github.com/eco/eco-routes/commit/4ed1aa148492a91f7ada5a25b20a35ddccf40d09), rather than tightening it to require an empty vault. This is an intentional policy change: child identity and salt uniqueness belong to the builder. Distinct parents may deliberately reuse the same child; Inbox still rejects replay of the same parent. Adding funds does not increase the child's declared reward or delivery obligation.

Prefunding of any amount is now allowed. We still measure the pre-transfer balance to verify the **new transfer's delta**, so existing funds cannot mask an outbound short transfer. Terminal children remain rejected, including with `publish=false`.

Tests cover below/equal/above-push prefunding, publication disabled, distinct-parent child reuse, same-parent replay rejection, and one declared reward payout followed by surplus refund. NatSpec and README no longer claim that the vault is unknowable or cannot be prefunded. This supersedes the earlier collision-rejection policy discussed above.

re1ro:
@MSevey addressed the requested regression and remaining nits in [4ed1aa1](https://github.com/eco/eco-routes/commit/4ed1aa148492a91f7ada5a25b20a35ddccf40d09). Ready for re-review.

- Added real `Portal.fulfill` / `Executor` tests with a token that charges a fee only when the chainer sends. They assert the complete `CallFailed` wrapper containing the exact `PushShortfall(vault, expected, delivered)`, not an inbound minimum-amount failure.
- Cases cover an empty vault, a prefunded vault, publication disabled, and a pre-existing chainer balance. Assertions cover solver input, swap inventory, chainer/vault and Executor balances, total supply, and parent claimant rollback.
- Two Hardhat tests also mine failing transactions with publication enabled/disabled and verify status 0, empty receipt logs, no surviving `IntentPublished` / `IntentChained` events, and restored balances/claimant state. This checks actual receipts rather than relying on an execution-log recorder that can include reverted frames.
- Corrected optional-publication/address-resolution NatSpec, made zero/codeless Portals return `InvalidPortal` before measurement, and aligned both deployment scripts on `INTENT_CHAINER_V5`.

Also addressed Foivos's output-calculation suggestion and replied inline on prefunding. The chosen policy now permits prefunded child vaults; builder-owned child identity replaces the earlier collision rejection. The transfer-delta and terminal-state checks remain. README clarifies opaque route deadlines and caller-selected token donations; the rendering loop documents its bounded concatenation cost.

Validation: **819 Foundry tests across 43 suites, 114 Hardhat tests, and 9 Jest tests passed**. Touched-file formatting/lint, shell syntax, and whitespace checks passed.

Deployment boundary: **V4 remains deployed and unchanged; V5 is not deployed.** The V4 manifest is retained as historical evidence. Adopting this revision requires a later chainer-only rollout and client address updates, not a Portal or Vault redeployment.

Foivos (approved):

